PR #3, opened 2019-04-16
Bakdata Kafka Streams Helm Repository

fapaul:
This PR adds a public helm repository to the bakdata org.


charts/streams-app/templates/_helpers.tpl
```diff
@@ -0,0 +1,33 @@
+{{/* vim: set filetype=mustache: */}}
```
lawben: i'd prefer not having editor-specific stuff in our code :)

PR #4, opened 2019-05-09
Add default log properties

lawben:


src/main/resources/log4j.properties
```diff
@@ -0,0 +1,6 @@
+log4j.rootLogger=WARN, console
+log4j.logger.com.bayer=INFO
```
fapaul: Remove customer specific logger property

PR #12, opened 2019-06-28
Add unique AppID method

lawben:
Each stream application needs a unique value within the entire Kafka cluster. Usually, this could be set to: className-inputTopic-outputTopic.


src/main/java/com/bakdata/common_kafka_streams/KafkaStreamsApplication.java
```diff
@@ -103,6 +103,12 @@ public void run() {
 
     public abstract void buildTopology(StreamsBuilder builder);
 
+    /**
+     * This must be set to a unique value within the entire Kafka cluster.
```
fapaul:
```suggestion
     * This must be set to a unique value for every application interacting with your kafka cluster to ensure internal state encapsulation. 
```

PR #21, opened 2019-08-07
Add JMX prometheus

lawben:


charts/streams-app/templates/jmx-configmap.yaml
```diff
@@ -0,0 +1,19 @@
...
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+data:
+  jmx-kafka-streams-app-prometheus.yml: |+
+    jmxUrl: service:jmx:rmi:///jndi/rmi://localhost:{{ .Values.jmx.port }}/jmxrmi
+    lowercaseOutputName: true
+    lowercaseOutputLabelNames: true
+    ssl: false
+    rules:
+    - pattern: ".*"
```
fapaul: We should make this configurable. In case someone wants to limit the metric space.

charts/index.yaml
```diff
@@ -2,12 +2,12 @@ apiVersion: v1
 entries:
   streams-app:
   - apiVersion: v1
-    created: 2019-07-18T14:12:27.254566+02:00
+    created: "2019-07-26T11:49:51.505279+02:00"
```
fapaul: Nit: Maybe we should rebuild the chart before every merge.

PR #22, opened 2019-08-09
Allow kafka streams configuration with external parameter

fapaul:


src/main/java/com/bakdata/common_kafka_streams/KafkaStreamsApplication.java
```diff
@@ -79,6 +81,9 @@
                     + "might cause inconsistent processing with multiple replicas.")
     private boolean forceReprocessing = false;
 
+    @CommandLine.Option(names = "--streams-config", split = " ", description = "Additional Kafka Streams properties")
```
philipp94831: Please use `split=","`

PR #24, opened 2019-08-16
Add rclone chart

lawben:


charts/rclone-copy/templates/include-pattern-config.yaml
```diff
@@ -0,0 +1,11 @@
+# This file is needed to work around the limitation that we cannot use 
+# the character '*' in the config YAML files, as this is a special character.
+# Howver, this is hidden from the values.yaml. We only need to specify a string
```
fapaul:
```suggestion
# However, this is hidden from the values.yaml. We only need to specify a string 
```

charts/rclone-copy/rclone.conf.template
```diff
@@ -0,0 +1,14 @@
+# Standard rclone config file. See https://rclone.org/docs/ for more details.
```
fapaul: Please delete this file.

PR #26, opened 2019-10-02
Change reprocessing to clean up

torbsto:
Starting the application with the cleanUp flag, all corresponding state dirs, offsets und internal topics are deleted. However, it does not reprocess the data and exists when it is done.

### DIFF
--- a/src/main/java/com/bakdata/common_kafka_streams/KafkaStreamsApplication.java
+++ b/src/main/java/com/bakdata/common_kafka_streams/KafkaStreamsApplication.java
@@ -78,11 +78,15 @@ public abstract class KafkaStreamsApplication implements Runnable, AutoCloseable
     @CommandLine.Option(names = {"-h", "--help"}, usageHelp = true, description = "print this help and exit")
     private boolean helpRequested = false;
 
-    @CommandLine.Option(names = "--cleanUp", arity = "1",
+    @CommandLine.Option(names = "--clean-up", arity = "1",
             description = "Clear the state store and the global Kafka offsets for the "
                     + "consumer group. Be careful with running in production and with enabling this flag - it "
                     + "might cause inconsistent processing with multiple replicas.")
     private boolean cleanUp = false;
+
+    @CommandLine.Option(names = "--delete-output", arity = "1",
+            description = "Delete the output topic during the clean up. ")
+    private boolean deleteOutputTopic = false;
 
     @CommandLine.Option(names = "--streams-config", split = ",", description = "Additional Kafka Streams properties")
     private Map<String, String> streamsConfig = new HashMap<>();
@@ -235,11 +239,15 @@ public abstract class KafkaStreamsApplication implements Runnable, AutoCloseable
         Runtime.getRuntime().addShutdownHook(new Thread(this::close));
     }
 
+    /**
+     * This methods resets the offset for all input topics and deletes internal topics, application state, and
+     * optionally the output topic.
+     */
     protected void runCleanUp() {
         this.inputTopics.stream()
                 .filter(topic -> !topic.isBlank())
                 .forEach(topic -> runResetter(topic, this.brokers, this.getUniqueAppId()));
-        if (!this.outputTopic.isBlank()) {
+        if (this.deleteOutputTopic && !this.outputTopic.isBlank()) {
             this.deleteOutputTopic();
         }
         this.streams.cleanUp();
--- a/src/main/java/com/bakdata/common_kafka_streams/KafkaStreamsApplication.java
+++ b/src/main/java/com/bakdata/common_kafka_streams/KafkaStreamsApplication.java
@@ -76,11 +76,11 @@ public abstract class KafkaStreamsApplication implements Runnable, AutoCloseable
     @CommandLine.Option(names = {"-h", "--help"}, usageHelp = true, description = "print this help and exit")
     private boolean helpRequested = false;
 
-    @CommandLine.Option(names = "--reprocess", arity = "1",
-            description = "Reprocess all data by clearing the state store and the global Kafka offsets for the "
+    @CommandLine.Option(names = "--cleanUp", arity = "1",
+            description = "Clear the state store and the global Kafka offsets for the "
                     + "consumer group. Be careful with running in production and with enabling this flag - it "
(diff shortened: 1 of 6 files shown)


src/test/java/com/bakdata/common_kafka_streams/WordCountTest.java
```diff
@@ -45,11 +45,11 @@
 
     @Test
     void shouldAggregateSameWordStream() {
-        this.testTopology.input().add("bla")
+        this.testTopology.input("Input").add("bla")
```
philipp94831: Did it not work before?

src/test/java/com/bakdata/common_kafka_streams/integration/CleanUpTest.java
```diff
@@ -0,0 +1,173 @@
...
+                        Arrays.asList("a", "b", "c")).useDefaults();
+        this.kafkaCluster.send(sendRequest);
+
+        this.runAndAssertSize(3);
+        this.runAndAssertSize(3);
+
+        // Wait until all stream application are completely stopped before triggering cleanup
+        Thread.sleep(TimeUnit.SECONDS.toMillis(TIMEOUT_SECONDS));
+        this.runCleanUp();
+        this.runAndAssertSize(6);
```
philipp94831: Why should there be 6 records after cleaning up and not sending any records?

src/test/java/com/bakdata/common_kafka_streams/integration/CleanUpTest.java
```diff
@@ -0,0 +1,173 @@
...
+
+    @Test
+    void shouldReprocessAlreadySeenRecords() throws InterruptedException {
+        final SendValuesTransactional<String> sendRequest =
+                SendValuesTransactional.inTransaction(this.app.getInputTopic(),
+                        Arrays.asList("a", "b", "c")).useDefaults();
+        this.kafkaCluster.send(sendRequest);
+
+        this.runAndAssertSize(3);
+        this.runAndAssertSize(3);
```
philipp94831: Why can I read the expected records twice?

src/test/java/com/bakdata/common_kafka_streams/WordCountTest.java
torbsto: It is necessary for now, but we probably should fix that in fluent-kafka-streams-test.
I gave the count store a name in the WordCount topology to make it more obvious that we use one. However, that changes the corresponding repartition topic. Without an explicit name, the store is named `KSTREAM-AGGREGATE-STATE-STORE-0000000003` and the topic `KSTREAM-AGGREGATE-STATE-STORE-0000000003-repartition`. Fluent filters `KSTREAM-*` topics so that they are not included in the input and output topics. With the store named `counts`, the topic is named `counts-repartiton` and therefore included in the input and output topic collection. Thus, I have to specify them explicitly.

src/test/java/com/bakdata/common_kafka_streams/integration/CleanUpTest.java
torbsto: The cleanUp doesn't affect the output topic.  It contains already 3 elements from the run before. After the cleanUp, the application reprocesses all elements from the input topic and writes them to the output topic.

src/test/java/com/bakdata/common_kafka_streams/integration/CleanUpTest.java
torbsto: The second run doesn't (re-)process any records since the app's offsets are known

src/test/java/com/bakdata/common_kafka_streams/integration/CleanUpTest.java
philipp94831: So the input topic is not cleaned?

torbsto: The input topic's content is not deleted, but the offset is reset.

philipp94831: Ok but we need to clear either the input or output topic

philipp94831: Clearing the output topic would be perfect. In case we change something internally we would consume the input from the beginning and only the new data is in the output topic

src/test/java/com/bakdata/common_kafka_streams/WordCountTest.java
SvenLehmann: Isn't this obsolete with https://github.com/bakdata/fluent-kafka-streams-tests/pull/34?

src/test/java/com/bakdata/common_kafka_streams/integration/CleanUpTest.java
```diff
@@ -0,0 +1,240 @@
...
+                        new KeyValue<>("blub", 2L)
+                );
+
+        this.runAndAssertContent(this.app, this.app.getOutputTopic(), expectedValues,
+                "WordCount contains all elements after first run");
+        this.runAndAssertContent(mirror, mirror.getOutputTopic(), expectedValues,
+                "Mirror contains all elements after first run");
+
+        this.runCleanUpWithDeletion(this.app);
+        // recreate deleted topic
```
philipp94831: Add an assert that the topic does not exist

src/main/java/com/bakdata/common_kafka_streams/KafkaStreamsApplication.java
```diff
@@ -76,11 +78,15 @@
...
-            description = "Reprocess all data by clearing the state store and the global Kafka offsets for the "
+    @CommandLine.Option(names = "--clean-up", arity = "1",
+            description = "Clear the state store and the global Kafka offsets for the "
                     + "consumer group. Be careful with running in production and with enabling this flag - it "
                     + "might cause inconsistent processing with multiple replicas.")
-    private boolean forceReprocessing = false;
+    private boolean cleanUp = false;
+
+    @CommandLine.Option(names = "--delete-output", arity = "1",
+            description = "Delete the output topic during the clean up. ")
```
philipp94831: trailing space

PR #32, opened 2019-10-30
 Delete schemas during clean up 

torbsto:


src/main/java/com/bakdata/common_kafka_streams/KafkaStreamsApplication.java
```diff
@@ -257,6 +261,13 @@ protected void runCleanUp() {
             }
         }
         this.streams.cleanUp();
+
+        try {
+            this.resetSchemaRegistry();
```
philipp94831: only do that if deleteOutput is set. Therefore, I suggest that you add this method call to deleteTopic()

src/main/java/com/bakdata/common_kafka_streams/KafkaStreamsApplication.java
```diff
@@ -271,6 +282,16 @@ protected DeleteTopicsResult deleteTopic(final String topic) {
         }
     }
 
+    private void resetSchemaRegistry() throws IOException, RestClientException {
+        final SchemaRegistryClient client = new CachedSchemaRegistryClient(this.schemaRegistryUrl, 100);
+        for (final String type : List.of("-key", "-value")) {
+            for (final String subject : this.inputTopics) {
```
philipp94831: the producing application is responsible for schemas of input topics. Thus, you can remove this part

src/test/java/com/bakdata/common_kafka_streams/integration/CleanUpTest.java
```diff
@@ -164,6 +172,46 @@ void shouldReprocessAlreadySeenRecords() throws InterruptedException {
...
+        final TestRecord testRecord = TestRecord.newBuilder().setContent("val 1").build();
+        final SendValuesTransactional<TestRecord> sendRequest = SendValuesTransactional
+                .inTransaction(this.app.getInputTopic(), Collections.singletonList(testRecord))
+                .with("schema.registry.url", this.schemaRegistryMockExtension.getUrl())
+                .with(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG, KafkaAvroSerializer.class.getName())
+                .build();
+
+        this.kafkaCluster.send(sendRequest);
+        this.app.run();
+        this.softly.assertThat(client.getAllSubjects()).hasSize(1);
```
philipp94831: please be more specific and make assertions such as assertThat(client.getAllSubjects()).contains("my-topic-value"). Also add a test were another schema is still in the schema registry after clean up, i.e., the schema registry is not wiped completely

src/main/java/com/bakdata/common_kafka_streams/KafkaStreamsApplication.java
```diff
@@ -248,15 +252,18 @@ protected void runCleanUp() {
         this.inputTopics.stream()
                 .filter(topic -> !topic.isBlank())
                 .forEach(topic -> runResetter(topic, this.brokers, this.getUniqueAppId()));
+
         if (this.deleteOutputTopic) {
             if (!this.outputTopic.isBlank()) {
                 this.deleteTopic(this.outputTopic);
+                this.resetSchemaRegistry();
```
philipp94831: this needs to be done for the error topic as well. Please move it to the deleteTopic() method

PR #37, opened 2019-11-14
Make PodAffinity rule configurable

SvenLehmann:
This PR allows setting either a required or a preferred PodAffinity rule.


charts/streams-app/templates/deployment.yaml
```diff
@@ -30,14 +30,27 @@ spec:
     spec:
       affinity:
         podAntiAffinity:
+        {{- if .Values.antiAffinityRequiredEnabled }}
```
philipp94831: I don't like this variable name... Also what is the default value?

SvenLehmann: Renamed the variable. The default value is `false`, so the 'preferred anti affinity' is used.

PR #38, opened 2019-11-15
Classify all Kafka errors as recoverable bar some exceptions

philipp94831:
Most Kafka errors are either recoverable or should stop processing. Therefore, they should be forwarded by default by all error handlers.

Furthermore, the clean up only logged errors upon schema registry clean up. This behavior has been changed to a thrown exception.


src/main/java/com/bakdata/common_kafka_streams/util/ErrorUtil.java
```diff
@@ -49,29 +50,58 @@
...
-     * @return whether the exception represents a schema registry timeout or not
+     * @return whether exception is thrown by Kafka and recoverable or not
+     */
+    public static boolean isRecoverableKafkaError(final Exception e) {
+        if (ORG_APACHE_KAFKA_COMMON_ERRORS.equals(e.getClass().getPackageName())) {
+            if (e instanceof RecordTooLargeException) {
+                return false;
+            }
+            if (e instanceof SerializationException) {
+                return e.getCause() instanceof SocketTimeoutException;
```
SvenLehmann: Add a comment

PR #39, opened 2019-11-15
Add S3 backed Serde

philipp94831:
This PR adds a special Serde that allows serialization of large messages by storing them on Amazon S3


philipp94831:
Moved to bakdata/kafka-s3-backed-serde

PR #40, opened 2019-11-26
Reset internal topics

torbsto:
This PR adds resetting subjects of internal topics. It iterates over the topology to get the corresponding names. Additionally, it extracts the clean up in a separate class and adds the deletion of manually created topics, that are not output or error topic (e.g. used for `through`)


src/main/java/com/bakdata/common_kafka_streams/CleanUpRunner.java
```diff
@@ -0,0 +1,159 @@
...
+        this.client = new CachedSchemaRegistryClient(application.getSchemaRegistryUrl(), 100);
+        this.inputTopics = application.getInputTopics();
+        this.brokers = application.getBrokers();
+        this.deleteOutputTopic = application.isDeleteOutputTopic();
+        this.streams = application.getStreams();
+    }
+
+    public void run() {
+        runResetter(String.join(",", this.inputTopics), this.brokers, this.streamsId);
+        if (this.deleteOutputTopic) {
```
philipp94831: This can be a method parameter

src/main/java/com/bakdata/common_kafka_streams/CleanUpRunner.java
```diff
@@ -0,0 +1,159 @@
...
+import org.apache.kafka.streams.KafkaStreams;
+import org.apache.kafka.streams.Topology;
+import org.apache.kafka.streams.TopologyDescription.Node;
+import org.apache.kafka.streams.TopologyDescription.Processor;
+import org.apache.kafka.streams.TopologyDescription.Sink;
+
+
+@Slf4j
+public class CleanUpRunner {
+    private final Topology topology;
```
philipp94831: Add `@NonNull`

src/main/java/com/bakdata/common_kafka_streams/CleanUpRunner.java
```diff
@@ -0,0 +1,159 @@
...
+                "--application-id", appId,
+                "--bootstrap-servers", brokers,
+                "--input-topics", inputTopics
+        };
+        final StreamsResetter resetter = new StreamsResetter();
+        resetter.run(args);
+    }
+
+    protected void deleteTopics() {
+        final List<Node> nodes = this.getNodes(this.topology);
```
philipp94831: I would move the getNodes method etc. to another class. Also: does getNodes return input topics? We do not want to delete those in any case

src/main/java/com/bakdata/common_kafka_streams/CleanUpRunner.java
```diff
@@ -0,0 +1,159 @@
...
+    private final Topology topology;
+    private final String streamsId;
+    private final Properties kafkaProperties;
+    private final SchemaRegistryClient client;
+    private final List<String> inputTopics;
+    private final String brokers;
+    private final boolean deleteOutputTopic;
+    private final KafkaStreams streams;
+
+    public CleanUpRunner(final KafkaStreamsApplication application) {
```
philipp94831: Use Builder and move instance creation to KafkaStreamsApplication

src/test/java/com/bakdata/common_kafka_streams/integration/CleanUpTest.java
```diff
@@ -215,6 +214,65 @@ void shouldDeleteKeySchema(final SoftAssertions softly)
                 .contains(this.app.getInputTopic() + "-key");
     }
 
+    @Test
```
philipp94831: Please add a test for a topology with a non-internal topic in between input and output

src/test/java/com/bakdata/common_kafka_streams/integration/CleanUpTest.java
```diff
@@ -175,9 +178,7 @@ void shouldDeleteValueSchema(final SoftAssertions softly)
                 .build();
 
         this.kafkaCluster.send(sendRequest);
-        this.app.run();
-        Thread.sleep(TimeUnit.SECONDS.toMillis(TIMEOUT_SECONDS));
-        this.app.close();
+        this.runApp();
```
philipp94831: why is close not needed anymore?

torbsto: It's still needed. I just extracted the three lines into a method.

philipp94831: My bad

src/test/java/com/bakdata/common_kafka_streams/integration/CleanUpTest.java
torbsto: There is a non-internal topic used by `through` in the `ComplexTopologyApplication`. Is there a different use case that should be tested?
I merged the schema deletion tests for non-output topics to make it a little clearer and  to reduce the run time.

philipp94831: That's totally fine, good job

PR #59, opened 2020-03-13
Allow specification of boolean flags without any value

philipp94831:


src/main/java/com/bakdata/common_kafka_streams/KafkaStreamsApplication.java
```diff
@@ -81,16 +81,16 @@
     private String schemaRegistryUrl = "";
     @CommandLine.Option(names = "--productive", arity = "1")
```
torbsto: Missing change?

philipp94831: No because this is by default true

PR #61, opened 2020-03-23
Run streams resetter for external source topics and intermediate topics

philipp94831:


src/main/java/com/bakdata/common_kafka_streams/util/TopologyInformation.java
```diff
@@ -42,6 +43,37 @@ public TopologyInformation(final Topology topology, final String streamsId) {
...
+    private static Stream<String> getAllSinks(final Collection<Node> nodes) {
+        return nodes.stream()
+                .filter(node -> node instanceof Sink)
+                .map(node -> ((Sink) node))
+                .map(Sink::topic);
+    }
+
+    private static boolean isInternalTopic(final String topic) {
+        return topic.startsWith("KSTREAM-") || topic.startsWith("KTABLE-")
+                || topic.endsWith("-repartition") || topic.endsWith("-changelog");
```
torbsto: assuming all topics with those suffixes being internal doesn't seem right. 
For `"-changelog",` you can check if it is a backing topic (i.e. `getBackingTopics(nodes).contains(topic)`). `"-repartition"` would require an analogous method to `getBackingTopics`.

philipp94831: Hm. In what case do we consider a topic ending with -repartition not internal?

torbsto: Depending on the definition of internal, I would argue that topics used in `through` are not internal. But obviously they can end with -repartition.

philipp94831: What do you think now?

PR #62, opened 2020-03-24
Override log4j.properties when using provided

BJennWare:
log4j.properties can be provided for helm with --set-file FILE_PATH


charts/streams-app/templates/log4j-configmap.yaml
```diff
@@ -0,0 +1 @@
+{{ if .Values.log4jConfig }}apiVersion: v1kind: ConfigMapmetadata:  name: {{ template "streams-app.name" . }}-log4j-configmapdata:  log4j.properties: |{{ .Values.log4jConfig | indent 4 }}{{ end }}
```
philipp94831: Is that supposed to be just one line?

philipp94831:
@BJennWare Please also make the changes to the statefulset

charts/streams-app/templates/log4j-configmap.yaml
BJennWare: This is strange. This file has 9 lines on my file system. I will push the file with linux line endings again

charts/streams-app/templates/log4j-configmap.yaml
```diff
@@ -0,0 +1,9 @@
+{{ if .Values.log4jConfig }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ template "streams-app.name" . }}-log4j-configmap
+data:
+  log4j.properties: |
+{{ .Values.log4jConfig | indent 4 }}
```
philipp94831: What is the purpose of this? Can I still configure log4j.properties via values.yaml? Or just mount a file?

BJennWare: We cannot just mount a file into helm. If we use helm --set-file log4jConfig=FILE_PATH than the contents of the file are in .Values.log4jConfig.

TLDR: We change the log config by mounting a log4j.properties file with --set-file

philipp94831: Ok

PR #64, opened 2020-04-07
Use Kubernetes secrets for password parameters

yannick-roeder:


charts/streams-app/templates/deployment.yaml
```diff
@@ -103,6 +103,13 @@ spec:
             - name: "{{ .Values.configurationEnvPrefix }}_ERROR_TOPIC"
               value: {{ .Values.streams.errorTopic | quote }}
           {{- end }}
+          {{- range $key, $value := .Values.secrets }}
+            - name: "{{ $key }}"
+              valueFrom:
+                secretKeyRef:
+                  name: "{{ $key | lower}}"
+                  key: value
```
philipp94831, line 111: What if I have multiple secrets? Are there conflicts if it is always "value"?

charts/streams-app/templates/deployment.yaml
```diff
@@ -103,6 +103,13 @@ spec:
             - name: "{{ .Values.configurationEnvPrefix }}_ERROR_TOPIC"
               value: {{ .Values.streams.errorTopic | quote }}
           {{- end }}
+          {{- range $key, $value := .Values.secrets }}
+            - name: "{{ $key }}"
```
philipp94831, line 107: quote? (see line 118)

charts/streams-app/templates/deployment.yaml
```diff
@@ -103,6 +103,13 @@ spec:
...
+              valueFrom:
+                secretKeyRef:
+                  name: "{{ $key | lower}}"
```
philipp94831, line 110: Why lower? Quote?

charts/streams-app/templates/deployment.yaml
```diff
@@ -161,3 +168,13 @@ spec:
                 path: log4j.properties
         {{- end }}
       {{- end }}
+{{- range $key, $value := .Values.secrets }}
```
philipp94831, line 171: Please create a separate secrets.yaml

charts/streams-app/templates/deployment.yaml
yannick-roeder: No, I don't think so, but I didn't test it. You could also use it for saving `username` and `password` in the same secret. See here: https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-environment-variables

Should I test it shortly?

philipp94831: Sounds fine but testing is never a bad idea :)

charts/streams-app/templates/deployment.yaml
```diff
@@ -161,3 +168,13 @@ spec:
...
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: "{{ $key | lower}}"
+type: Opaque
+stringData:
+  value: {{ $value }}
```
philipp94831, line 179: quote?

charts/streams-app/templates/deployment.yaml
yannick-roeder: Lower, because keys are only allowed to be lowercase and have `.` and `-` in it. The quotes don't change anything in the test: https://gitlab.bayer.com/sih/kafka-deployments/alignment-deployment/-/jobs/671556#L286

philipp94831: I feel safer with quote additionally

charts/streams-app/templates/deployment.yaml
yannick-roeder: `{{ $value | quote}}` or like `"{{ $value }}"`?

philipp94831: Whatever we did elsewhere

yannick-roeder: Apparently we use both in different places. For `name:` normally `"{{ $name }}"` and for `value:` usually `{{ $value | quote }}`. In the end both produces the same result.

charts/streams-app/templates/deployment.yaml
yannick-roeder:
```
ABC:  <set to the key 'value' in secret 'abc'>  Optional: false
FOO:  <set to the key 'value' in secret 'foo'>  Optional: false
```

Looks fine to me in the description of the pod.

charts/streams-app/templates/deployment.yaml
```diff
@@ -161,3 +168,13 @@ spec:
                 path: log4j.properties
         {{- end }}
       {{- end }}
+{{- range $key, $value := .Secrets.secrets }}
+---
```
philipp94831, line 172: Just make this a separate secrety.yaml and everything is fine :)

charts/streams-app/templates/values.yaml
```diff
@@ -0,0 +1,10 @@
+{{- range $key, $value := .Values.secrets }}
```
philipp94831, line 1: Wrong file name? Also, did you run helm package?

charts/streams-app/templates/deployment.yaml
```diff
@@ -103,6 +103,13 @@ spec:
             - name: "{{ .Values.configurationEnvPrefix }}_ERROR_TOPIC"
               value: {{ .Values.streams.errorTopic | quote }}
           {{- end }}
+          {{- range $key, $value := .Values.secrets }}
```
philipp94831, line 106: We need this also for streams-app-static and streams-cleanup-job

charts/streams-app/templates/secrets.yaml
```diff
@@ -0,0 +1,11 @@
+{{- if .Values.secrets }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ template "streams-app.fullname" . }}
+type: Opaque
+data:
+  {{- range $key, $value := .Values.secrets }}
+  {{ $key }}: {{ $value | b64enc }}
```
philipp94831, line 9: Do we need to do anything to be able to read the value as well?

charts/streams-app-cleanup-job/templates/secrets.yaml
```diff
@@ -0,0 +1,11 @@
+{{- if .Values.secrets }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ template "streams-app.fullname" . }}
```
philipp94831, line 5: streams-app-cleanup-job ?

PR #70, opened 2020-06-03
Fix yaml file for kubernetes > 1.16

VictorKuenstler:
Update deployment.yaml to work with kubernetes > 1.16


philipp94831:
Please also do this for the Streams clean up job

PR #73, opened 2020-06-11
Remove container suffix from container name

### LINKED ISSUE #69: Container can not be launched if name exceeds 63 characters
Kubernetes allows no more than 63 characters as the container name. The container name is composed of streams app name (truncated to 63 characters) and -container. Those 10 additional characters make the helm deployment fail. Container name should therefore be truncated additionally.

yannick-roeder:
Fixes #69


charts/index.yaml
```diff
@@ -2,31 +2,57 @@ apiVersion: v1
...
+    - rclone-copy-1.0.0.tgz
+    version: 1.0.0
+  - apiVersion: v1
+    created: 2020-10-26T09:49:59.9602532+01:00
     description: A helm chart for rclone copy on Kubernetes.
     digest: 76bfa8090f35f661898da7748513d4f454a75ab26507bc01f35b617762048d0f
     name: rclone-copy
     urls:
-    - rclone-copy/rclone-copy-1.0.0.tgz
+    - rclone-copy%5C/rclone-copy-1.0.0.tgz
```
philipp94831, line 18: This happens on windows but needs to be reverted (also other occurrences)

charts/index.yaml
```diff
@@ -2,7 +2,15 @@ apiVersion: v1
...
   - apiVersion: v1
-    created: 2020-09-15T16:23:59.6324708+02:00
+    created: 2020-10-26T09:49:59.9562577+01:00
+    description: A helm chart for rclone copy on Kubernetes.
+    digest: 92e2565243c7084d9a03e955cba3328106ebe39419e1f19549239ecac6d87ea3
+    name: rclone-copy
+    urls:
+    - rclone-copy-1.0.0.tgz
+    version: 1.0.0
+  - apiVersion: v1
```
philipp94831: This looks kinda duplicated. Can you remove index.yaml and create it from scratch again?

philipp94831 (approved):

charts/index.yaml
BJennWare: Should be fixed now

PR #75, opened 2020-08-04
Specify additional output topics via default CLI

philipp94831:


src/main/java/com/bakdata/common_kafka_streams/KafkaStreamsApplication.java
```diff
@@ -197,6 +199,10 @@ public String getInputTopic() {
         return this.getInputTopics().get(0);
     }
 
+    protected Optional<String> getOutputTopic(final String role) {
+        return Optional.ofNullable(this.extraOutputTopics.get(role));
```
philipp94831, line 203: What do you think @VictorKuenstler @BJennWare : Should we return Optional or throw an exception if an output topic does not exist? i.e., delegate error handling to caller or handle it by default

BJennWare: Can you think of a scenario for only outputting to a topic if configured? But having the freedom to do something like this might be worthwhile so I am fine with returning an optional.

philipp94831: It's more for a case where it hasn't been configured. The question is if we would like to have  a default error message then or let the user write one

BJennWare: If it is only about an error message we can have a default one that includes the topic name and the role.

philipp94831: Better now?

PR #82, opened 2020-11-06
Add scaled object for autoscaling with keda

yannick-roeder:


charts/streams-app/values.yaml
```diff
@@ -50,6 +50,21 @@ jmx:
...
 
+autoscaling:
+  enabled: false
+  # consumergroup: foo
+  lagThreshold: "1000"
+  pollingInterval: 30
+  cooldownPeriod: 300
+  offsetResetPolicy: earliest
+  minReplicas: 0
+  maxReplicas: 50
```
VictorKuenstler, line 61: What do you think about also commenting this out? I think max replicas should not have a default value to no unintentionally have 50 replicas by accident, causing problems in smaller clusters.

yannick-roeder: Good point

charts/streams-app/values.yaml
```diff
@@ -50,6 +50,21 @@ jmx:
   metricRules:
     - ".*"
 
+autoscaling:
+  enabled: false
+  # consumergroup: foo
+  lagThreshold: "1000"
```
yannick-roeder, line 56: How about this value? Is it ok to set a default here?

VictorKuenstler: I think it is not that critical but should also be configured for every deployment, right?

charts/streams-app/templates/scaled-object.yaml
```diff
@@ -0,0 +1,33 @@
...
+  name: {{ template "streams-app.fullname" . }}-so
+spec:
+  scaleTargetRef:
+    name: {{ template "streams-app.fullname" . }}
+  pollingInterval: {{ .Values.autoscaling.pollingInterval }}
+  cooldownPeriod:  {{ .Values.autoscaling.cooldownPeriod }}
+  minReplicaCount: {{ .Values.autoscaling.minReplicas }}
+  maxReplicaCount: {{ .Values.autoscaling.maxReplicas }}
+  triggers:
+    {{- range .Values.streams.inputTopics }}
```
philipp94831, line 15: Please also iterate extraInputTopics (it is a map, so only check the values of the map)

charts/streams-app/templates/scaled-object.yaml
```diff
@@ -0,0 +1,33 @@
...
+    - type: kafka
+      metadata:
+        bootstrapServers: {{ $root.Values.streams.brokers }}
+        consumerGroup: {{ $root.Values.autoscaling.consumergroup }}
+        topic: {{ . | quote }}
+        lagThreshold: {{ $root.Values.autoscaling.lagThreshold | quote }}
+        offsetResetPolicy: {{ $root.Values.autoscaling.offsetResetPolicy }}
+    {{- end }}
+    {{- range .Values.autoscaling.topics }}
```
philipp94831, line 24: Can you merge the topics first in order to loop only once?

philipp94831 (changes requested):

charts/streams-app/values.yaml
yannick-roeder: Yes, it might be different for every deployment

charts/streams-app/templates/scaled-object.yaml
yannick-roeder: Do we eve need the `autoscaling.topics` then?

philipp94831: Yes, there are internal topics automatically created by Kafka or created by us for repartitioning

yannick-roeder: Only leaving a comment for merging for now, because `concat list1 list2` is not supported in helm 2

charts/streams-app/templates/scaled-object.yaml
```diff
@@ -0,0 +1,44 @@
...
+        topic: {{ . | quote }}
+        lagThreshold: {{ $root.Values.autoscaling.lagThreshold | quote }}
+        offsetResetPolicy: {{ $root.Values.autoscaling.offsetResetPolicy }}
+    {{- end }}
+    {{- range $key, $value := .Values.streams.extraInputTopics }}
+    - type: kafka
+      metadata:
+        bootstrapServers: {{ $root.Values.streams.brokers }}
+        consumerGroup: {{ $root.Values.autoscaling.consumergroup }}
+        topic: {{ . | quote }}
```
philipp94831, line 40: that should be $value, right?

yannick-roeder: Yes, in my test it worked like this, but I'm going to change it

philipp94831 (approved):

PR #83, opened 2020-11-09
Fix comment for autoscaling

yannick-roeder:


charts/streams-app/values.yaml
```diff
@@ -61,7 +61,7 @@ autoscaling:
   # maxReplicas: 50
   ## all topics from streams.inputTopics are automatically taken, so
```
philipp94831, line 62: you should also mention streams.extraInputTopics

charts/streams-app/values.yaml
```diff
@@ -61,7 +61,7 @@ autoscaling:
...
   ## only use the 'additionalTopics' option for adding more topics
```
philipp94831, line 63: comment should also say `topics`. Also put an `, e.g., internal topics` here

philipp94831 (changes requested):

philipp94831 (approved):

PR #87, opened 2020-12-01
Add tolerations to deployment

yannick-roeder:


charts/streams-app/templates/deployment.yaml
```diff
@@ -46,6 +46,12 @@ spec:
         {{ $key }}: {{ $value }}
         {{- end }}
     spec:
+      {{- if .Values.tolerations.key }}
```
philipp94831, line 49: tolerations should be a list. Maybe you need to do it similar to resources where you just put the yaml as is

yannick-roeder: Should we still support default values for `operator` and `effect`, so that we can do something like:
```
tolerations:
  - key1
  - key2
      operator: Exists
      effect: NoExecute
  - key3
```

philipp94831: Just let the user put there what ever he would like in order to be fully flexible wrt tolreations. No defaults or sth like that

charts/rclone-copy/templates/rclone-cron.yaml
```diff
@@ -33,6 +33,10 @@ spec:
             {{ $key }}: {{ $value }}
             {{- end }}
         spec:
+          {{- if .Values.tolerations }}
+          tolerations:
+{{ toYaml .Values.tolerations | indent 12 }}
```
philipp94831, line 38: Did you check if the indents are correct for all 3 charts?

yannick-roeder: Yes, now I checked all of them

philipp94831 (approved):

PR #89, opened 2020-12-17
Add producer application

philipp94831:


charts/producer-app/templates/_helpers.tpl
```diff
@@ -0,0 +1,33 @@
...
+{{/*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+If release name contains chart name it will be used as a full name.
+*/}}
+{{- define "producer-app.fullname" -}}
+{{- if .Values.fullnameOverride -}}
+{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
+{{- else -}}
+{{- $name := default .Chart.Name .Values.nameOverride -}}
```
yannick-roeder, line 18: What are those additions in this and the following lines for?

charts/producer-app/templates/cron-job.yaml
```diff
@@ -0,0 +1,89 @@
...
+            {{- end }}
+            {{- if and (hasKey .Values.streams "extraOutputTopics") (.Values.streams.extraOutputTopics) }}
+              - name: "{{ .Values.configurationEnvPrefix }}_EXTRA_OUTPUT_TOPICS"
+                value: "{{- range $key, $value := .Values.streams.extraOutputTopics }}{{ $key }}={{ $value }},{{- end }}"
+            {{- end }}
+            {{- range $key, $value := .Values.secrets }}
+              - name: "{{ $key }}"
+                valueFrom:
+                  secretKeyRef:
+                    name: {{ template "producer-app.fullname" $ }}
```
yannick-roeder, line 76: Not sure, but don't you need to use `$root` here instead of `$`?

charts/producer-app/templates/_helpers.tpl
philipp94831: I don't now where I copied it from originally but this helm chart basically moved here from our customer's GitLab

yannick-roeder: Ok, because in every other helper file we don't have these extra steps here

charts/producer-app/templates/cron-job.yaml
philipp94831: We don't do that for streams app either

src/main/java/com/bakdata/common_kafka_streams/util/SchemaTopicClient.java
```diff
@@ -0,0 +1,121 @@
...
+     */
+    public static CachedSchemaRegistryClient createSchemaRegistryClient(@NonNull final Properties configs,
+            @NonNull final String schemaRegistryUrl) {
+        final Map<String, Object> originals = new HashMap<>();
+        configs.forEach((key, value) -> originals.put(key.toString(), value));
+        return new CachedSchemaRegistryClient(schemaRegistryUrl, 100, originals);
+    }
+
+    /**
+     * Delete a topic if it exists and reset the delete the corresponding schema registry topics.
```
yannick-roeder, line 80: Typo in description

yannick-roeder:
Did you test the deployments somewhere? Please also make sure that your tests succeed

philipp94831:
> Did you test the deployments somewhere? Please also make sure that your tests succeed

works @yannick-roeder

b-feldmann (approved):

PR #91, opened 2021-01-14
Upgrade to helm 3

yannick-roeder:


charts/producer-app-cleanup-job/templates/job.yaml
```diff
@@ -13,7 +13,7 @@ metadata:
     {{- end }}
 spec:
   ttlSecondsAfterFinished: 30
-  replicas: {{ .Values.replicaCount }}
+  parallelism: {{ .Values.replicaCount }}
```
philipp94831, line 16: Replicas for our jobs is actually wrong, just remove the variable (also check the streams job)

philipp94831 (changes requested):

charts/producer-app-cleanup-job/templates/job.yaml
yannick-roeder: Helm 3 is validating the openapi now before deploying so that something like this results in an error

philipp94831 (approved):

PR #97, opened 2021-02-10
Add priority class to helm charts

yannick-roeder:


charts/producer-app-cleanup-job/templates/job.yaml
```diff
@@ -35,6 +35,9 @@ spec:
       {{- if .Values.tolerations }}
       tolerations:
 {{ toYaml .Values.tolerations | indent 8 }}
+      {{- end }}
+      {{- if .Values.priorityClassName }}
```
philipp94831: Not needed for clean up jobs

philipp94831 (approved):

charts/producer-app-cleanup-job/templates/job.yaml
yannick-roeder: I think you forgot to refresh

philipp94831: I did and the comment disappeared because the line disappeared, but magically it still got submitted^^

PR #98, opened 2021-02-12
Update Kafka to 2.7

philipp94831:


src/main/java/com/bakdata/kafka/KafkaStreamsApplication.java
```diff
@@ -171,10 +171,7 @@ protected String getInputTopic(final String role) {
      * @see KafkaStreams#setUncaughtExceptionHandler(UncaughtExceptionHandler)
      */
     protected UncaughtExceptionHandler getUncaughtExceptionHandler() {
-        return (t, e) -> {
-            log.debug("Closing resources because of uncaught exception");
-            this.closeResources();
-        };
+        return null;
```
philipp94831, line 174: This is now always called when StateListener is called and leads to two invocations of closeResources(). I changed it therefore back to how it was some time ago. What do you think?

torbsto: Do you know why it was needed before? I'd have expected that the state listener on-error callback is always called if there had been an uncaught exception?

torbsto: Generally, it might be cleaner to return a no-op handler instead of `null`? Just for backward compatibility

philipp94831: Null means nothing happens https://github.com/bakdata/streams-bootstrap/blob/master/src/main/java/com/bakdata/kafka/KafkaStreamsApplication.java#L108 (method should return Optional but I don't want to change the API). Don't know why it wasn't called before. I have the feeling this changes with every Kafka release...

torbsto (approved):

PR #104, opened 2021-05-06
Reset schema of internal topics when they are deleted 

torbsto:
Previously, a clean-up run reset the schema of internal topics only when `--delete-output` was set. The topics themselves however were deleted even without the flag.
With this PR, the flag is no longer required.

### DIFF
--- a/src/test/java/com/bakdata/kafka/integration/StreamsCleanUpTest.java
+++ b/src/test/java/com/bakdata/kafka/integration/StreamsCleanUpTest.java
@@ -26,7 +26,7 @@ package com.bakdata.kafka.integration;
 
 
 import static net.mguenther.kafka.junit.EmbeddedKafkaCluster.provisionWith;
-import static net.mguenther.kafka.junit.EmbeddedKafkaClusterConfig.useDefaults;
+import static net.mguenther.kafka.junit.EmbeddedKafkaClusterConfig.defaultClusterConfig;
 
 import com.bakdata.kafka.KafkaStreamsApplication;
 import com.bakdata.kafka.TestRecord;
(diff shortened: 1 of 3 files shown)


src/test/java/com/bakdata/kafka/integration/StreamsCleanUpTest.java
```diff
@@ -362,18 +390,43 @@ void shouldDeleteSchemaOfIntermediateTopics(final SoftAssertions softly)
         softly.assertThat(client.getAllSubjects())
                 .contains(inputSubject, internalSubject, backingSubject, manualSubject);
 
-        this.runCleanUpWithDeletion();
+        this.runCleanUp();
```
philipp94831, line 393: Through topics are not affected in this case, right? Only true internal topics (ending on -repartition e.g.)

torbsto: Exactly, that's why added separate tests for intermediate topics (the ones used in `through()`/`repartition()`)

philipp94831: I was not sure about that because in the diff only two tests use `runCleanUp()`

philipp94831: `shouldDeleteInternalTopics` is the test that covers the case?

torbsto: yes, `shouldDeleteInternalTopics` and `shouldDeleteSchemaOfInternalTopics` cover clean up runs without the `--delete-output` flag, i.e. make sure that internal topics are completly deleted but intermediate topics are not touched.
`shouldDeleteIntermediateTopics` and `shouldDeleteSchemaOfIntermediateTopics` ensure that intermediate topics and their schemas are deleted when the flag is set.

philipp94831 (approved):

PR #108, opened 2021-07-20
Make producer-app's CronJob schedule optional

torbsto:
In some instances (particularly demos), we don't need a CronJob but a regular Job. This commit makes the schedule optional. If it is not specified, the generated deployment will be of kind 'Job'.


charts/producer-app/templates/job.yaml
```diff
@@ -1,6 +1,6 @@
 {{- $root := . -}}
```
torbsto, line 1: I've also renamed this file since `job.yaml` fits both cases. Hope that's okay.

yannick-roeder: Yes, I think this should not cause any problems

yannick-roeder:
LGTM, I will ask @VictorKuenstler to have a look as well

VictorKuenstler (approved):
LGTM

yannick-roeder (approved):

PR #110, opened 2021-07-27
producer chart: job/cronjob switch

mkcode92:
* make producer-app deployable as job
* problem: when switching between job/cronjob a lot of shared config is needed in a different path (spec.template vs spec.jobTemplate.spec.template)
* this block is now separated as podTemplate


charts/producer-app/templates/job.yaml
```diff
@@ -11,86 +16,16 @@ metadata:
...
+  suspend: {{ .Values.suspend }}
   {{- if .Values.schedule }}
   schedule: {{ quote .Values.schedule }}
-  {{- end }}
-  suspend: {{ .Values.suspend }}
   successfulJobsHistoryLimit: 1
   concurrencyPolicy: Replace
   jobTemplate:
     spec:
       template:
```
yannick-roeder, line 26: Did you test this in the cluster? This is going to be an empty key when no schedule is set

yannick-roeder:
LGTM, did you test it in the cluster?

torbsto:
@VictorKuenstler  @philipp94831 This is a fix for the changes introduced by #108. I've mistakenly assumed that Job and CronJob have the same spec besides `schedule`. Obviously, the changes in this PR are more complex than those in #108. I still think it's worth to have it.

@mkcode92 I think you also have to update the `producer-app` chart according to the steps listed here: https://github.com/bakdata/streams-bootstrap/tree/master/charts#development

yannick-roeder:
> @mkcode92 I think you also have to update the `producer-app` chart according to the steps listed here: https://github.com/bakdata/streams-bootstrap/tree/master/charts#development

True, thanks @torbsto

mkcode92:
> LGTM, did you test it?

Yes, both should work!

yannick-roeder (approved):

yannick-roeder (approved):

PR #115, opened 2021-08-20
Migrate to Log4j 2

philipp94831:


src/test/resources/log4j2.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Configuration>
+  <Appenders>
+    <Console name="Console" target="SYSTEM_OUT">
+      <PatternLayout pattern="%d{HH:mm:ss.SSS} [%t] %-5level %logger{36} - %msg%n"/>
+    </Console>
+  </Appenders>
+  <Loggers>
+    <Logger name="com.bakdata" level="debug" additivity="false">
```
yannick-roeder, line 9: This was info before. Is debug intended here?

src/test/resources/log4j2.xml
```diff
@@ -0,0 +1,16 @@
...
+      <AppenderRef ref="Console"/>
+    </Logger>
+    <Root level="info">
```
yannick-roeder, line 12: This was warning before. Is info intended here?

src/test/resources/log4j2.xml
philipp94831: There was no test logging config before. Only production but that was never used

yannick-roeder: Ok, I thought it is the replacement for `log4j.logger.com.bakdata=INFO`

src/main/resources/log4j.properties
```diff
@@ -1,5 +0,0 @@
-log4j.rootLogger=WARN, console
```
yannick-roeder, line 1: What happened to this logging config then if the other one is a test config?

philipp94831: As I said: This was never used anywhere because this is a library and not a docker artifact

yannick-roeder (approved):

PR #121, opened 2021-09-09
Update openbridge/ob_bulkstash version and allow arguments

aqchen:


charts/rclone-copy/templates/rclone-cron.yaml
```diff
@@ -52,6 +52,9 @@ spec:
             args:
               - copy
               - -v
+            {{-  range .Values.arguments }}
```
philipp94831, line 55: If you change this, then please also add a corresponding field to the values. Probably an empty array but you should provide an example with comments. Or remove this if it is not needed

aqchen: It's not needed this time. But should we keep it for the future, since allowing arguments is useful?

philipp94831: If it is something useful then feel free to add it but also update the values (and don't forget to package :-) )

philipp94831 (approved):

PR #130, opened 2022-02-11
Add option for streams state persistence

philipp94831:


charts/streams-app/templates/deployment.yaml
```diff
@@ -220,3 +226,20 @@ spec:
...
+  volumeClaimTemplates:
+    - metadata:
+        name: datadir
+      spec:
+        accessModes: [ "ReadWriteOnce" ]
+        resources:
+          requests:
+            storage: "{{ .Values.persistence.size }}"
+        {{- if .Values.persistence.storageClass }}
+        {{- if (eq "-" .Values.persistence.storageClass) }}
```
yannick-roeder, line 239: In which case would the storage class be `-` or why would I set it to `-` specifically?

yannick-roeder: Nevermind, you commented it below

PR #135, opened 2022-03-01
Verify deletion of topics

### LINKED ISSUE #134: Clean up should check if topic deletion was successful
If there is an active consumer for a topic we attempt to delete, the topic is transitioned to a state with only one partition but still exists. Therefore, the clean up should list the topics after deletion and check if the topic still exists. If so, the clean up should fail

philipp94831:
Fixes #134

### DIFF
--- a/src/main/java/com/bakdata/kafka/util/TopicClient.java
+++ b/src/main/java/com/bakdata/kafka/util/TopicClient.java
@@ -111,6 +111,9 @@ public final class TopicClient implements Closeable {
         } catch (final ExecutionException | TimeoutException ex) {
             throw new KafkaAdminException("Failed to delete topic " + topicName, ex);
         }
+        if (this.exists(topicName)) {
+            throw new IllegalStateException("Deletion of topic " + topicName + "failed");
+        }
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/main/java/com/bakdata/kafka/util/TopicClient.java
```diff
@@ -111,6 +112,7 @@ public void deleteTopic(final String topicName) {
         } catch (final ExecutionException | TimeoutException ex) {
             throw new KafkaAdminException("Failed to delete topic " + topicName, ex);
         }
+        Preconditions.checkState(!this.exists(topicName), "Deletion of topic %s failed", topicName);
```
torbsto, line 115: This isn't a precondition. I'd either use Google's `Verify` or a plain if statement + exception.

philipp94831: Could you elaborate on the difference? Internally it just checks the condition and then throws

torbsto: see [JavaDoc for Precondition](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/base/Preconditions.html), under "Non-preconditions":

> It is of course possible to use the methods of this class to check for invalid conditions which are not the caller's fault. Doing so is **not recommended** because it is misleading to future readers of the code and of stack traces. See Conditional failures explained in the Guava User Guide for more advice. Notably, [Verify](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/base/Verify.html) offers assertions similar to those in this class for non-precondition checks.

torbsto (approved):

PR #141, opened 2022-05-04
Fix key-value pairs indentation inside pod.yaml file

irux:


philipp94831:
You need to package the chart as well

philipp94831:
The title of the PR is not a correct sentence

philipp94831 (approved):

PR #144, opened 2022-06-02
Update documentation

philipp94831:


README.md
```diff
@@ -77,36 +82,94 @@ public class StreamsBootstrapApplication extends KafkaStreamsApplication {
...
-For the configuration and deployment to Kubernetes, you can use the [Helm Charts](https://github.com/bakdata/streams-bootstrap/tree/master/charts).
-To configure your streams-bootstrap app, you can use the [`values.yaml`](https://github.com/bakdata/streams-bootstrap/blob/master/charts/streams-app/values.yaml) as a starting point. 
+For the configuration and deployment to Kubernetes, you can use
+the [Helm Charts](https://github.com/bakdata/streams-bootstrap/tree/master/charts).
+To configure your streams app, you can use
+the [`values.yaml`](https://github.com/bakdata/streams-bootstrap/blob/master/charts/streams-app/values.yaml) as a
+starting point.
+To configure your producer app, you can use
+the [`values.yaml`](https://github.com/bakdata/streams-bootstrap/blob/master/charts/producer-app/values.yaml) as a
+starting point.
```
VictorKuenstler, line 172: Maybe also add a hint to both cleanup jobs?

charts/streams-app/README.md
```diff
@@ -0,0 +1,110 @@
...
+| `streams.schemaRegistryUrl`  | URL of schema registry to connect to.                                                                                                                                            |         |
+| `streams.staticMembership`   | Whether to use [Kafka Static Group Membership](https://cwiki.apache.org/confluence/display/KAFKA/KIP-345%3A+Introduce+static+membership+protocol+to+reduce+consumer+rebalances). | `false` |
+| `streams.config`             | Configurations for your [Kafka Streams app](https://kafka.apache.org/documentation/#streamsconfigs).                                                                             | `{}`    |
+| `streams.inputTopics`        | List of input topics for your streams application.                                                                                                                               | `[]`    |
+| `streams.extraInputTopics`   | Map of additional named input topics.                                                                                                                                            | `{}`    |
```
VictorKuenstler, line 52: I think a context could help here: Used to name input topics for their usage, e.g. when joining different streams?

README.md
```diff
@@ -77,36 +86,103 @@ public class StreamsBootstrapApplication extends KafkaStreamsApplication {
...
+- `--extra-input-topics`: Additional named input topics if you need to specify multiple topics with different message
+  types (`<String=String>[,<String=String>...]`)
+
+- `--extra-input-patterns`: Additional named input patterns if you need to specify multiple topics with different
+  message types (`<String=String>[,<String=String>...]`)
+
+- `--extra-output-topics`: Additional named output topics if you need to specify multiple topics with different message
+  types (`String=String>[,<String=String>...]`)
+
+- `--clean-up`: Whether the state of the Kafka Streams app, i.e., offsets and state stores and auto-created topics,
```
VictorKuenstler, line 114:
```suggestion
- `--clean-up`: Whether the state of the Kafka Streams app, i.e., offsets and state stores and internal topics,
```

philipp94831: I wanted to highlight that it only affects auto-created topics. A topic that the user creates but is only used internally could also fall under your description

VictorKuenstler: Ah, good point 👍

charts/streams-app/README.md
```diff
@@ -0,0 +1,110 @@
...
+| `imageTag`         | Docker image tag of the Kafka Streams app. | `latest`     |
+| `imagePullPolicy`  | Docker image pull policy.                  | `Always`     |
+| `imagePullSecrets` | Secrets to be used for private registries. |              |
+
+### Streams
+
+| Parameter                    | Description                                                                                                                                                                      | Default |
+|------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------|
+| `streams.brokers`            | Comma separated list of Kafka brokers to connect to.                                                                                                                             |         |
+| `streams.schemaRegistryUrl`  | URL of schema registry to connect to.                                                                                                                                            |         |
```
yannick-roeder, line 48: I like these tables as an overview. I think an additional column `required` could be useful similar to how we did it [here](https://github.com/bakdata/ci-templates/tree/main/actions/helm-gke-deploy#input-parameters), because not every parameter with an empty default is also required (e.g. `priorityClassName`)

philipp94831: Besides the image basically nothing is required because everything depends on your scenario and you always need a subset of configs so I don't think it makes much sense here

yannick-roeder: Ok, I thought the `brokers` and `schemaRegistryUrl` are always required, then it is fine

charts/streams-app-cleanup-job/README.md
```diff
@@ -0,0 +1,20 @@
+# streams-app cleanup helm chart
+
+This chart can be used to deploy a cleanup job for your Kafka Streams app developed using streams-bootstrap.
+
```
mkcode92, line 4: Maybe you can repeat here, which steps cleanup involves. Also, you could elaborate on the intended use: should the app deployment be deleted before running the job? Which parts of the config must be the same as in the cleaned app?

charts/producer-app-cleanup-job/README.md
```diff
@@ -0,0 +1,12 @@
+# producer-app cleanup helm chart
+
+This chart can be used to deploy a cleanup job for your Kafka producer app developed using streams-bootstrap.
+
```
mkcode92, line 4: Maybe elaborate a bit, [cf.](#r888737919)

philipp94831: What would you expect here?

charts/streams-app-cleanup-job/README.md
philipp94831: I added a note. However, I won't repeat what the clean up does. That is already part of the global README and this is just about deploying it to Kubernetes

charts/producer-app-cleanup-job/README.md
mkcode92: Only what you just added...

yannick-roeder (dismissed):

VictorKuenstler (approved):

charts/producer-app/README.md
```diff
@@ -0,0 +1,64 @@
...
+| `imagePullSecrets` | Secrets to be used for private registries.  |               |
+
+### Streams
+
+| Parameter                   | Description                                                                                                | Default |
+|-----------------------------|------------------------------------------------------------------------------------------------------------|---------|
+| `streams.brokers`           | Comma separated list of Kafka brokers to connect to.                                                       |         |
+| `streams.schemaRegistryUrl` | URL of schema registry to connect to.                                                                      |         |
+| `streams.config`            | Configurations for your [Kafka producer app](https://kafka.apache.org/documentation/#producerconfigs).     | `{}`    |
+| `streams.outputTopic`       | Output topic for your producer application.                                                                |         |
```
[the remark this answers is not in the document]
philipp94831: You can use extra output topic. This is totally up to the user. Also this is just the deployment config. It depends on the implementation of your streams app.

charts/streams-app/README.md
```diff
@@ -0,0 +1,110 @@
...
+| `streams.extraInputPatterns` | Map of additional named input patterns if you need to specify multiple topics with different message types.                                                                      | `{}`    |
+| `streams.outputTopic`        | Output topic for your streams application.                                                                                                                                       |         |
+| `streams.extraOutputTopics`  | Map of additional named output topics if you need to specify multiple topics with different message types.                                                                       | `{}`    |
+| `streams.errorTopic`         | Error topic for your streams application.                                                                                                                                        |         |
+| `streams.productive`         | Whether to use Kafka configuration values that are more suitable for production environments.                                                                                    | `true`  |
```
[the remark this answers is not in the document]
philipp94831: This is just the deployment config. Please read the docs for stremas-bootstrap library

raminqaf (approved):

PR #150, opened 2022-11-16
Support multiple extra input topics

philipp94831:


src/main/java/com/bakdata/kafka/KafkaStreamsApplication.java
```diff
@@ -80,8 +80,10 @@ public abstract class KafkaStreamsApplication extends KafkaApplication implement
     protected Pattern inputPattern;
     @CommandLine.Option(names = "--error-topic", description = "Error topic")
     protected String errorTopic;
-    @CommandLine.Option(names = "--extra-input-topics", split = ",", description = "Additional named input topics")
-    protected Map<String, String> extraInputTopics = new HashMap<>();
+    @CommandLine.Option(names = "--extra-input-topics", split = ",", description = "Additional named input topics",
+            converter = {StringConverter.class, StringListConverter.class})
+    // PicoCLI does not allow generics for map values
```
torbsto, line 85: Pretty sure that it works. You can use `converter = {UseDefaultConverter.class, StringListConverter.class}` and let `StringListConverter` implement `ITypeConverter<List<String>>`

torbsto: See also this test: https://github.com/remkop/picocli/blob/2add8987874925fa1b52f7f172c8388bb3f8771a/src/test/java/picocli/GenericTest.java#L96

philipp94831: I am confused. This is obviously (except the UseDefault) what I tried and it raised an error inside PicoCLI which only allowed Optionals as generics. I don't know where my mistake was but yeah it is apparently working

philipp94831: I thought I also tested v4.7...

torbsto (approved):

PR #153, opened 2022-11-29
Clean schema registry subjects of named foreign key joins

philipp94831:


src/main/java/com/bakdata/kafka/util/TopologyInformation.java
```diff
@@ -45,6 +45,8 @@
  * Representation of the nodes of a Kafka Streams topology
  */
 public class TopologyInformation {
+    private static final String SUBSCRIPTION_REGISTRATION_SUFFIX = "-subscription-registration-topic";
+    private static final String SUBSCRIPTION_RESPONSE_SUFFIX = "subscription-response-topic";
```
raminqaf, line 49: a minimal notice: shouldn't this contain a `-` at the beginning?
like: `-subscription-response-topic`

raminqaf (approved):

PR #173, opened 2023-02-21
Fix casing of consumerGroup value

disrupted:


charts/streams-app/values.yaml
```diff
@@ -23,29 +23,29 @@ resources:
     memory: 2G
 
 streams:
-#  brokers: "test:9092"
-#  schemaRegistryUrl: "url:1234"
+  # brokers: "test:9092"
```
philipp94831, line 26: Isn't the indentation wrong now? When uncommenting it, the values should be correct

disrupted: previously they were weirdly formatted with the hash position at the beginning of the line. compare to autoscaling config below which is formatted correctly

philipp94831 (approved):

PR #178, opened 2023-04-12
Support deployment of bootstrap application with an HTTP server

torbsto:
This PR adds support for the Helm chart deployment of bootstrap application (both producer and streams application) that run an HTTP server by:
- allowing the definition of ports
- support of readiness and liveness probes
- opt-in creation of services

I did a test deployment for all features, but probably not all feature combinations.


charts/streams-app/README.md
```diff
@@ -22,6 +22,12 @@ Alternatively, a YAML file that specifies the values for the parameters can be p
...
+| `ports.schema`           | Protocol for port. Must be UDP, TCP, or SCTP (optional).                                                                                                                                                                                                   |                                            |
+| `ports.servicePort`      | Number of the port of the service (optional). See [service definition](#service)                                                                                                                                                                           |                                            |
+| `livenessProbe`          | Probe v1 definition for producer: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#probe-v1-core                                                                                                                                       | `{}`                                       |
```
philipp94831, line 29: Is producer the correct term here?

philipp94831 (approved):

PR #180, opened 2023-04-17
Fix service selector to match pod labels

raminqaf:
This PR fixes an issue related to the [new support of the services](https://github.com/bakdata/streams-bootstrap/pull/178) in Streams Bootstrap.
The service selector did not match the pod(s) labels. Therefore the service was not able to find the pod(s). So [the service has no Endpoints](https://kubernetes.io/docs/tasks/debug/debug-application/debug-service/#does-the-service-have-any-endpoints).

### DIFF
--- a/charts/producer-app/templates/service.yaml
+++ b/charts/producer-app/templates/service.yaml
@@ -13,7 +13,10 @@ metadata:
 spec:
   selector:
     app: {{ template "producer-app.name" . }}
-    chart: {{ template "producer-app.chart" . }}
+    release: {{ .Release.Name }}
+    {{- range $key, $value := .Values.labels }}
+    {{ $key }}: {{ $value }}
+    {{- end }}
   ports:
   {{- range .Values.ports }}
     {{- if .servicePort }}
--- a/charts/producer-app/templates/service.yaml
+++ b/charts/producer-app/templates/service.yaml
@@ -14,9 +14,6 @@ spec:
   selector:
     app: {{ template "producer-app.name" . }}
     release: {{ .Release.Name }}
-    {{- range $key, $value := .Values.labels }}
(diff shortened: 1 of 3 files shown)


charts/producer-app/templates/service.yaml
```diff
@@ -13,7 +13,10 @@ metadata:
 spec:
   selector:
     app: {{ template "producer-app.name" . }}
-    chart: {{ template "producer-app.chart" . }}
+    release: {{ .Release.Name }}
+    {{- range $key, $value := .Values.labels }}
+    {{ $key }}: {{ $value }}
+    {{- end }}
```
yannick-roeder, line 19: Does a selector need to have all pod labels or would `app` and `release` already be enough?

charts/producer-app/values.yaml
```diff
@@ -1,5 +1,7 @@
 nameOverride: bakdata-producer-app
 
+replicaCount: 1
+
```
yannick-roeder, line 4: It makes sense to set a default of 1, but please remove the comment in lines 11-12 then

yannick-roeder (changes requested):

charts/producer-app/templates/service.yaml
raminqaf: Agree with you on that. I just set it to see what feedback comes back.

charts/producer-app/values.yaml
torbsto: Does it make sense? It only applies if deployment is enabled and that isn't the default. 

If you keep it, it should be updated in the docs

philipp94831: As this is unrelated to the actual change in this PR, maybe just leave it as it was

yannick-roeder (approved):

PR #188, opened 2023-09-22
Add modules for easy usage with Fluent Kafka Streams Tests and Large Message SerDe

philipp94831:


azure-pipelines.yml
```diff
@@ -19,6 +19,7 @@ resources:
     - repository: templates
       type: github
       name: bakdata/bakdata-project-templates
+      ref: tmp/upload-snapshot
```
philipp94831, line 22: TODO: revert

lombok.config
```diff
@@ -1,4 +1,4 @@
 # This file is generated by the 'io.freefair.lombok' Gradle plugin
 config.stopBubbling = true
 lombok.addLombokGeneratedAnnotation = true
-lombok.extern.findbugs.addSuppressFBWarnings = true
+#lombok.extern.findbugs.addSuppressFBWarnings = true FIXME
```
philipp94831, line 4: TODO

streams-bootstrap-large-messages/src/main/java/com/bakdata/kafka/LargeMessageKafkaStreamsApplication.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+public abstract class LargeMessageKafkaStreamsApplication extends KafkaStreamsApplication {
+
+    @Override
+    protected void cleanUpRun(final CleanUpRunner cleanUpRunner) {
+        this.registerLargeMessageCleanUpHook(cleanUpRunner);
+        super.cleanUpRun(cleanUpRunner);
+    }
+
+    private void registerLargeMessageCleanUpHook(final CleanUpRunner cleanUpRunner) {
```
torbsto, line 40: Can we move this into the utility class? Then it wouldn't be necessary to use this subclass, which can be hard in some cases where you already have a custom super class

build.gradle.kts
```diff
@@ -16,6 +11,7 @@ allprojects {
 
     tasks.withType<Test> {
         maxParallelForks = 1 // Embedded Kafka does not reliably work in parallel since Kafka 3.0
+        useJUnitPlatform()
```
raminqaf, line 14:
```suggestion
tasks {
    test {
        useJUnitPlatform()
    }
}
```

streams-bootstrap-test/src/main/java/com/bakdata/kafka/StreamsBootstrapTopologyFactory.java
```diff
@@ -0,0 +1,111 @@
...
+     * KafkaStreamsApplication}
+     *
+     * @see KafkaStreamsApplication#getKafkaProperties()
+     * @see KafkaStreamsApplication#createTopology()
+     */
+    public static <K, V> TestTopologyExtension<K, V> createTopologyExtension(final KafkaStreamsApplication app) {
+        return new TestTopologyExtension<>(app::createTopology, app.getKafkaProperties());
+    }
+
+    private static Topology createTopology(final KafkaStreamsApplication app, final Properties properties) {
```
raminqaf, line 106: Just to make reading from the top of the class a bit easier. Maybe rename the method to this:
```suggestion
    private static Topology createTopologyWithSRConfigured(final KafkaStreamsApplication app, final Properties properties) {
```

streams-bootstrap-test/src/main/java/com/bakdata/kafka/StreamsBootstrapTopologyFactory.java
```diff
@@ -0,0 +1,111 @@
...
+        app.setSchemaRegistryUrl(properties.getProperty(AbstractKafkaSchemaSerDeConfig.SCHEMA_REGISTRY_URL_CONFIG));
```
raminqaf, line 107: Can we use the [SR's mock protocol](https://github.com/confluentinc/schema-registry/blob/764e345f1f943065801d5034f285c3ad4156eaaf/client/src/main/java/io/confluent/kafka/schemaregistry/testutil/MockSchemaRegistry.java#L39) as Torben suggested?

raminqaf: Something like this:
```suggestion
        app.setSchemaRegistryUrl("mock://test-scope");
```

philipp94831: No, this is about using the SR mock provided with fluent Kafka

philipp94831: We should rather evaluate whether we can get rid of SR mock in fluent Kafka altogether

build.gradle.kts
philipp94831: Why?

streams-bootstrap-large-messages/src/main/java/com/bakdata/kafka/LargeMessageKafkaStreamsApplication.java
philipp94831: Done. Do you think it makes sense to even have this super class?

streams-bootstrap-test/src/main/java/com/bakdata/kafka/StreamsBootstrapTopologyFactory.java
philipp94831: Refactored a little bit but done

streams-bootstrap/build.gradle.kts
```diff
@@ -0,0 +1,46 @@
+description = "Collection of commonly used modules when writing a Kafka Streams Application"
+
+plugins {
+    id("com.github.davidmc24.gradle.plugin.avro") version "1.5.0"
+}
+
+dependencies {
+    val kafkaVersion: String by project
+    implementation(group = "org.apache.kafka", name = "kafka_2.13", version = kafkaVersion)
```
torbsto, line 9: @raminqaf mentioned this a while back: Why do we need this as dependency?

streams-bootstrap-large-messages/build.gradle.kts
```diff
@@ -0,0 +1,6 @@
+description = "Utils for using Large Message SerDe with your Kafka Streams Application"
+
+dependencies {
+    api(project(":streams-bootstrap"))
+    implementation(group = "com.bakdata.kafka", name = "large-message-core", version = "2.5.1")
```
torbsto, line 5: Could also be `api` given that you most likely need to access the config

torbsto (approved):

streams-bootstrap/build.gradle.kts
philipp94831: I think it is because of StreamsResetter

streams-bootstrap-large-messages/build.gradle.kts
philipp94831: What do you prefer? That way the user needs to explicitly specify a version of LMS

philipp94831: And also this dependency does not include the Serde which the user most likely needs to add

torbsto (approved):

PR #193, opened 2023-11-22
Fix usage of `nameOverride`

### LINKED ISSUE #185: Investigate the necessity of `nameOverride`
Having to set the `nameOverride` for every deployment seems like a workaround. We should look into setting the correct Helm release name automatically without overriding.

raminqaf:
Fixes #185


philipp94831:
@raminqaf Is this still needed? I think it works pretty good now with KPOps

raminqaf:
Closing this because the issue was solved with KPOps

PR #197, opened 2024-01-26
Add --unique-app-id parameter

philipp94831:


streams-bootstrap/src/main/java/com/bakdata/kafka/KafkaStreamsApplication.java
```diff
@@ -159,7 +161,9 @@ public void close() {
      * This must be set to a unique value for every application interacting with your kafka cluster to ensure internal
      * state encapsulation. Could be set to: className-inputTopic-outputTopic
      */
-    public abstract String getUniqueAppId();
+    public String getUniqueAppId() {
```
raminqaf: If the user doesn't implement this and calls it, they will get `null` returned, correct? 🤔

philipp94831: See comments in Slack/new version

streams-bootstrap/src/main/java/com/bakdata/kafka/KafkaStreamsApplication.java
```diff
@@ -271,13 +267,31 @@ protected Properties createKafkaProperties() {
...
-        kafkaConfig.setProperty(StreamsConfig.APPLICATION_ID_CONFIG, this.getUniqueAppId());
+        kafkaConfig.setProperty(StreamsConfig.APPLICATION_ID_CONFIG, this.getApplicationId());
 
         this.configureDefaultSerde(kafkaConfig);
         kafkaConfig.setProperty(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, this.getBrokers());
         return kafkaConfig;
     }
 
+    private String getApplicationId() {
+        final String uniqueAppId = this.getUniqueAppId();
```
raminqaf, line 278: Does this method exist? I cannot find it anywhere 🤔

streams-bootstrap/src/main/java/com/bakdata/kafka/KafkaStreamsApplication.java
```diff
@@ -271,13 +267,31 @@ protected Properties createKafkaProperties() {
...
+        if (uniqueAppId == null) {
+            if (this.uniqueAppId == null) {
+                throw new IllegalArgumentException("Must pass --application-id or implement #getUniqueAppId()");
```
raminqaf, line 281: You removed the abstract method. How can the user implement the `getUniqueAppId()`?

philipp94831: The field is covered by Lombok `@Getter`, tests still override it

philipp94831: We can also change the field name and do not generate a Getter for it 🤔

raminqaf: It was just slightly confusing that the flag was --application-id and the field name was `uniqueAppId`

streams-bootstrap/src/main/java/com/bakdata/kafka/KafkaStreamsApplication.java
```diff
@@ -271,13 +273,31 @@ protected Properties createKafkaProperties() {
...
         // topology
-        kafkaConfig.setProperty(StreamsConfig.APPLICATION_ID_CONFIG, this.getUniqueAppId());
+        kafkaConfig.setProperty(StreamsConfig.APPLICATION_ID_CONFIG, this.getApplicationId());
 
         this.configureDefaultSerde(kafkaConfig);
         kafkaConfig.setProperty(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, this.getBrokers());
         return kafkaConfig;
     }
 
+    private String getApplicationId() {
```
torbsto, line 283: I guess it would clutter the API even more and lead to confusion together with `getUniqueAppId()`, but I feel like we will need to call it elsewhere

torbsto, line 283: And I think we need a test for this method

torbsto (commented):
I would slightly prefer a new parameter in `streams` that lets you set the consumer group instead of the flag, and I really don't see why it would break anything.

streams-bootstrap/src/main/java/com/bakdata/kafka/KafkaStreamsApplication.java
philipp94831: Will add tests, just wanted to have your thoughts first

streams-bootstrap/build.gradle.kts
```diff
@@ -49,3 +49,10 @@ dependencies {
 
     testImplementation(group = "com.ginsberg", name = "junit5-system-exit", version = "1.1.2")
 }
+
+tasks.withType<Test> {
+    jvmArgs(
+        "--add-opens=java.base/java.lang=ALL-UNNAMED",
+        "--add-opens=java.base/java.util=ALL-UNNAMED"
```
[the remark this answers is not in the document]
philipp94831: Outdated. Merged in #199. We use JDK 17 for building and that requires add-opens

charts/streams-app-cleanup-job/values.yaml
```diff
@@ -15,6 +15,7 @@ configurationEnvPrefix: "APP"
 streams:
 #  brokers: "test:9092"
 #  schemaRegistryUrl: "url:1234"
+  passApplicationId: true
```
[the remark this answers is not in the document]
philipp94831: New helm chart should have it as default but old apps can still use new chart. We did a similar thing for optimizeLeaveGroupBehavior

streams-bootstrap/src/main/java/com/bakdata/kafka/KafkaStreamsApplication.java
philipp94831: Hm, this method does the validation whenever Kafka properties are created. So getUniqueAppId() should still return the correct result unless a user explicitly overrides the method and it returns null. I don't think this will ever happen

torbsto:
@raminqaf @philipp94831 Like I said above, I'd prefer using a separate argument for the app id and not reusing the autoscaling one. See https://github.com/bakdata/streams-bootstrap/compare/feature/application-id-param...feature/helm-app-id for how I imagine it could work

philipp94831:
> @raminqaf @philipp94831 Like I said above, I'd prefer using a separate argument for the app id and not reusing the autoscaling one. See [feature/application-id-param...feature/helm-app-id](https://github.com/bakdata/streams-bootstrap/compare/feature/application-id-param...feature/helm-app-id) for how I imagine it could work

I feel like having two parameters is more confusing. Then we should document it well and probably deprecat the old way of configuring it

torbsto:
I would be fine with deprecating it, but I'm not sure if there might be a situation in which someone wants to use KEDA but not set the consumer group via Helm.
I also added this check so that we fail as early as possible:
https://github.com/bakdata/streams-bootstrap/blob/b86810e10570613f8ebcd2064b390b8e8c2920d6/charts/streams-app/templates/scaled-object.yaml#L2-L4

But I agree that we should also document it properly in the values.yaml and README.md

philipp94831:
> @raminqaf @philipp94831 Like I said above, I'd prefer using a separate argument for the app id and not reusing the autoscaling one. See [feature/application-id-param...feature/helm-app-id](https://github.com/bakdata/streams-bootstrap/compare/feature/application-id-param...feature/helm-app-id) for how I imagine it could work

There is also this annotation we should not forget about https://github.com/bakdata/streams-bootstrap/blob/master/charts/streams-app/templates/deployment.yaml#L20

philipp94831:
Closing in favor of #243

PR #212, opened 2024-04-30
Add terminationGracePeriodSeconds

vostres:


charts/streams-app/templates/deployment.yaml
```diff
@@ -48,7 +48,7 @@ spec:
         prometheus.io/scrape: "true"
         prometheus.io/port: {{ .Values.prometheus.jmx.port | quote }}
       {{- end }}
-      {{- range $key, $value := .Values.annotations }}
+      {{- range $key, $value := .Values.podAnnotations }}
```
philipp94831, line 51: Please add a default empty list to values.yaml

charts/streams-app/templates/deployment.yaml
```diff
@@ -58,7 +58,7 @@ spec:
       labels:
         app: {{ template "streams-app.name" . }}
         release: {{ .Release.Name }}
-        {{- range $key, $value := .Values.labels }}
+        {{- range $key, $value := .Values.podLabels }}
```
philipp94831, line 61: Please add a default empty list to values.yaml

charts/streams-app/templates/deployment.yaml
```diff
@@ -99,6 +99,7 @@ spec:
       {{- if .Values.priorityClassName }}
       priorityClassName: {{ .Values.priorityClassName }}
       {{- end }}
+      terminationGracePeriodSeconds: {{ .Values.terminationGracePeriodSeconds }}
```
philipp94831, line 102: Please add a default of 300 to values.yaml

charts/streams-app/templates/deployment.yaml
philipp94831: There is also a condition above where you need to check podAnnotations instead

charts/streams-app/templates/deployment.yaml
philipp94831: There is also a condition above where you need to check podLabels instead

philipp94831 (changes requested):

philipp94831:
Please merge master

philipp94831:
Please add podAnnotations/podLabels also to producer-app, producer-app-clean-up-job, streams-app-clean-up-job

charts/streams-app/templates/deployment.yaml
vostres: > There is also a condition above where you need to check podLabels instead

I don't think there is a condition for Values.labels

philipp94831: Sorry, you are right

vostres:
> Please add podAnnotations/podLabels also to producer-app, producer-app-clean-up-job, streams-app-clean-up-job

so in case there are already Values.annotations/Values.labels in chart under `spec.template.metadata` I should just refactor those and if there are none (as in producer-app) add them?

philipp94831:
Yeah, would be good to have proper annotation son producer app jobs/cronjobs

philipp94831:
We should split the PR in two: one for terminationGracePeriod, one for labels/annotations

vostres:
> We should split the PR in two: one for terminationGracePeriod, one for labels/annotations

I guess I than close this pull request and create two new ones?

philipp94831:
> > We should split the PR in two: one for terminationGracePeriod, one for labels/annotations
> 
> I guess I than close this pull request and create two new ones?

Just open one new one. You can use this one for one of the two things

philipp94831 (changes requested):
Please update README of the chart

philipp94831 (approved):

PR #216, opened 2024-05-07
Fix naming of helm releases

philipp94831:


disrupted (commented):
do we need to update any instructions for the default Helm usage (without KPOps) now that the `-clean` suffix isn't appended automatically anymore?

philipp94831:
> do we need to update any instructions for the default Helm usage (without KPOps) now that the `-clean` suffix isn't appended automatically anymore?

I don't think so. Job will now have the name of the helm release but everything still should work

charts/producer-app-cleanup-job/templates/_helpers.tpl
```diff
@@ -17,7 +17,7 @@ If release name contains chart name it will be used as a full name.
 {{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
 {{- else -}}
 {{- $name := default .Values.nameOverride -}}
-{{- printf "%s-clean" $name | trunc 63 | trimSuffix "-" -}}
+{{- printf "%s" $name | trunc 63 | trimSuffix "-" -}}
```
disrupted, line 20: probably the formatstring is not needed anymore and there's a better way to write it, but I am fine with it

disrupted (approved):

charts/producer-app-cleanup-job/templates/_helpers.tpl
philipp94831: I copied it from other charts in this repo. You are probably right but I would keep it for now

PR #217, opened 2024-05-07
Add Values.podLabels and Values.podAnnotations

vostres:
Add Values.podLabels and Values.podAnnotations to:

producer-app

producer-app-clean-up-job

streams-app-clean-up-job


charts/producer-app-cleanup-job/templates/job.yaml
```diff
@@ -15,19 +15,19 @@ spec:
   ttlSecondsAfterFinished: 30
   template:
     metadata:
-    {{- if or .Values.awsRole .Values.annotations }}
```
philipp94831, line 18: annotations are not used anymore. Either we reuse them from the values or we add them to the job metadata (see above, labels are there). Any opinion @torbsto ?

charts/streams-app-cleanup-job/templates/job.yaml
```diff
@@ -15,19 +15,19 @@ spec:
   ttlSecondsAfterFinished: 30
   template:
     metadata:
-    {{- if or .Values.awsRole .Values.annotations }}
```
philipp94831, line 18: same comment as above

charts/producer-app/templates/pod.yaml
```diff
@@ -1,19 +1,19 @@
 {{- define "producer-app.podTemplate" -}}
 {{- $root := . -}}
 metadata:
-{{- if or .Values.awsRole .Values.annotations }}
```
philipp94831, line 4: same comment as above

philipp94831 (changes requested):
Please update READMEs

charts/producer-app-cleanup-job/templates/job.yaml
philipp94831: I think we should just add them for consistency

charts/producer-app/templates/pod.yaml
vostres: > same comment as above

producer-app already has Values.annotations in deployment.yaml. 
but there are only Values.labels in job.yaml. Should I add it there?

philipp94831: yes please

charts/streams-app/templates/deployment.yaml
```diff
@@ -48,17 +48,17 @@ spec:
         prometheus.io/scrape: "true"
         prometheus.io/port: {{ .Values.prometheus.jmx.port | quote }}
       {{- end }}
-      {{- range $key, $value := .Values.annotations }}
+      {{- range $key, $value := .Values.podAnnotations }}
         {{ $key | quote }}: {{ $value | quote }}
       {{- end }}
-      {{- if and .Values.autoscaling.consumerGroup (not .Values.annotations.consumerGroup) }}
+      {{- if and .Values.autoscaling.consumerGroup (not .Values.podAnnotations.consumerGroup) }}
```
philipp94831, line 54: We don't actually need the consumer group annotation on the pod, only on the deployment, so you can remove it here

philipp94831:
Can you please also update rclone-copy chart?

charts/streams-app/templates/deployment.yaml
```diff
@@ -48,7 +48,7 @@ spec:
         prometheus.io/scrape: "true"
         prometheus.io/port: {{ .Values.prometheus.jmx.port | quote }}
       {{- end }}
-      {{- range $key, $value := .Values.annotations }}
+      {{- range $key, $value := .Values.podAnnotations }}
         {{ $key | quote }}: {{ $value | quote }}
       {{- end }}
       {{- if and .Values.autoscaling.consumerGroup (not .Values.annotations.consumerGroup) }}
```
philipp94831, line 54: I meant removing this whole 3 line if statement :)

charts/streams-app/templates/deployment.yaml
```diff
@@ -39,7 +39,7 @@ spec:
       release: {{ .Release.Name }}
   template:
     metadata:
-    {{- if or (or (or .Values.awsRole .Values.prometheus.jmx.enabled) .Values.autoscaling.consumerGroup) .Values.annotations }}
+    {{- if or (or (or .Values.awsRole .Values.prometheus.jmx.enabled) .Values.autoscaling.consumerGroup) .Values.podAnnotations }}
```
philipp94831, line 42: remove the consumerGroup condition along with the next comment

philipp94831 (changes requested):

charts/producer-app-cleanup-job/templates/job.yaml
```diff
@@ -4,6 +4,12 @@ apiVersion: batch/v1
 kind: Job
 metadata:
   name: {{ template "producer-app-cleanup-job.fullname" . }}
+{{ - if .Values.annotations }}
```
philipp94831, line 7:
```suggestion
{{- if .Values.annotations }}
```

does that cause the issues?

charts/streams-app/templates/deployment.yaml
vostres: > remove the consumerGroup condition along with the next comment

I changed it, but here I am not 100% sure, if I got the condition right

philipp94831: Looks good! (and much much better with the single or)

philipp94831 (changes requested):
Please merge main

PR #224, opened 2024-05-15
Make affinity in charts more generic

yannick-roeder:


charts/streams-app/templates/deployment.yaml
```diff
@@ -60,50 +60,23 @@ spec:
...
-        {{- end }}
-        podAntiAffinity:
-        {{- if .Values.requirePodAntiAffinity }}
-          requiredDuringSchedulingIgnoredDuringExecution:
-          - labelSelector:
-            matchExpressions:
-            - key: app
-              operator: In
-              values:
-              - {{ template "streams-app.name" . }}
```
philipp94831, line 79: Since this is a dynamic value injected by helm, I don't see how we can replicate that with the new way

charts/producer-app-cleanup-job/templates/job.yaml
```diff
@@ -39,24 +38,19 @@ spec:
       serviceAccountName: {{ .Values.serviceAccountName }}
       {{- end }}
       {{- if .Values.tolerations }}
-      tolerations:
-{{ toYaml .Values.tolerations | indent 8 }}
+      tolerations: {{ toYaml .Values.tolerations | nindent 8 }}
```
philipp94831, line 41: Can you do these changes in a separate PR please?

philipp94831 (changes requested):

charts/streams-app/templates/deployment.yaml
torbsto: You can achieve the same if we use `tpl` in the deployment.yaml, e.g.:
```yaml
      {{- if .Values.affinity }}
      affinity: 
        {{- tpl (toYaml .Values.affinity) . | nindent 8 }}
      {{- end }}
```

and the values.yaml:
```yaml
  podAntiAffinity:
     preferredDuringSchedulingIgnoredDuringExecution:
      - weight: 1
        podAffinityTerm:
          topologyKey: "kubernetes.io/hostname"
          labelSelector:
            matchExpressions:
              - key: app
                operator: In
                values:
                  - '{{ .Values.nameOverride }}'
```

philipp94831 (changes requested):
Please update READMEs

philipp94831 (approved):

PR #233, opened 2024-06-27
fix: add gh actions badge

DerTiedemann:
fixes https://github.com/orgs/bakdata/projects/7/views/2?pane=issue&itemId=68163319


philipp94831:
The link in the description does not point to an issue

PR #238, opened 2024-07-19
Remove dependency on log4j

### LINKED ISSUE #162: Remove dependency on log4j
streams-bootstrap includes log4j as a dependency, because it configures the log level when `--debug` is set.
While this is a convenient feature, using this library with a different logging implementation is almost infeasible.
I'd suggest we simply remove this dependency. We could further offer a method that let users do custom configurations if debug is set.

philipp94831:
Fixes #162

### DIFF
--- a/streams-bootstrap-cli/src/main/java/com/bakdata/kafka/KafkaApplication.java
+++ b/streams-bootstrap-cli/src/main/java/com/bakdata/kafka/KafkaApplication.java
@@ -42,8 +42,6 @@ import lombok.RequiredArgsConstructor;
 import lombok.Setter;
 import lombok.ToString;
 import lombok.extern.slf4j.Slf4j;
-import org.apache.logging.log4j.Level;
-import org.apache.logging.log4j.core.config.Configurator;
 import picocli.CommandLine;
 import picocli.CommandLine.Command;
 import picocli.CommandLine.ParseResult;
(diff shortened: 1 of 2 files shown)


streams-bootstrap-cli/src/main/java/com/bakdata/kafka/KafkaApplication.java
```diff
@@ -299,20 +294,16 @@ public final CleanableApp<CR> createCleanableApp() {
...
 
     /**
-     * Configure application when running in debug mode. By default, Log4j2 log level is configured to debug for
-     * {@code com.bakdata} and the applications package.
+     * Called before starting the application, e.g., invoking {@link #run()}
      */
-    protected void configureDebug() {
-        Configurator.setLevel("com.bakdata", Level.DEBUG);
-        Configurator.setLevel(this.getClass().getPackageName(), Level.DEBUG);
+    protected void onApplicationStart() {
```
torbsto, line 299: I'm still not entirely sure when we opt for methods that we can overwrite (like `createTopicConfig`) and when for interfaces that can be passed (like the clean up hooks)

philipp94831: This is the base class for all cli apps and it is an abstract class so I opted for a method that can be overriden. Adding some kind of interface here makes things much more confusing I think. Have a field that is set via constructor of the abstract class? Return an instance of the interface instead of just calling the method? What other options are there?

torbsto (approved):

PR #240, opened 2024-07-22
Rename extra topics to labeled topics

### LINKED ISSUE #231: Rename extra-input/output topic to named-input/output topic

philipp94831:
Fixes #231


streams-bootstrap-cli/src/test/java/com/bakdata/kafka/CliTest.java
```diff
@@ -287,28 +287,28 @@ public void run() {
                     "--brokers", "brokers",
                     "--schema-registry-url", "schema-registry",
                     "--input-topics", "input1,input2",
-                    "--extra-input-topics", "role1=input3,role2=input4;input5",
+                    "--named-input-topics", "role1=input3,role2=input4;input5",
```
raminqaf, line 290: Do we want to change `role` to `name`?

philipp94831: Im the example we can. I. The Java API I was unsure because it can be kinda confused with topic name and every topic obviously has a name. Maybe we find better terms?

raminqaf: some suggestions:
- alias
- identifier
- tag
- label
- stream

torbsto: I think key could also become very confusing given their meaning in Kafka. I like label and tag, identifier is also ok but a bit too long :smile:

philipp94831: so labeled-input-topics etc?

JakobEdding (approved):

PR #242, opened 2024-07-22
Rename `--brokers` to `--bootstrap-servers`

### LINKED ISSUE #232: Rename `--brokers` to `--bootstrap-server`
This is used by all Kafka CLI tools

philipp94831:
Fixes #232


charts/streams-app/templates/deployment.yaml
```diff
@@ -115,9 +115,9 @@ spec:
             - name: "{{ .Values.configurationEnvPrefix }}_VOLATILE_GROUP_INSTANCE_ID"
               value: "true"
           {{- end }}
-          {{- if hasKey .Values.streams "brokers" }}
-            - name: "{{ .Values.configurationEnvPrefix }}_BROKERS"
-              value: {{ .Values.streams.brokers | quote }}
+          {{- if hasKey .Values.streams "bootstrapServers" }}
+            - name: "{{ .Values.configurationEnvPrefix }}_BOOTSTRAP_SERVER"
```
philipp94831, line 119: The cli param for all kafka cli tools is always `--bootstrap-server`. However it is `StreamsConfig.BOOTSTRAP_SERVERS_CONFIG`. Thats why I named it bootstrapServers in the chart. wdyt?

torbsto: I would prefer if we're consistent and then stick to using the property names everywhere, so `bootstrap.servers` in this case.

philipp94831: So you would use `--bootstrap-servers`?

torbsto: yes

philipp94831: I added an alias to have both available

README.md
```diff
@@ -100,7 +100,7 @@ public class StreamsBootstrapApplication extends KafkaStreamsApplication {
 
 The following configuration options are available:
 
-- `--brokers`: List of Kafka brokers (comma-separated) (**required**)
+- `--bootstrap-servers`, `--bootstrap-server`: List of Kafka bootstrap servers (comma-separated) (**required**)
```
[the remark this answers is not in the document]
philipp94831: Did you see this conversation? https://github.com/bakdata/streams-bootstrap/pull/242#discussion_r1687819037 I don't think it is confusing tbh. In the helm chart there is only one option (bootstrapServers) and the singular alias just exists to use the same as kafka CLI tools. Both have exactly the same effect

PR #258, opened 2024-10-28
Add annotations to all Kubernetes resources

raminqaf:


charts/streams-app/templates/_helpers.tpl
```diff
@@ -27,3 +27,20 @@ Create chart name and version as used by the chart label.
...
+  {{- range $key, $value := .Values.annotations }}
+    {{ $key | quote }}: {{ $value | quote }}
+  {{- end }}
+
+  {{- if and .Values.kafka.applicationId (not .Values.annotations.consumerGroup) }}
+    consumerGroup: {{ .Values.kafka.applicationId | quote }}
+  {{- end }}
+{{- end }}
+{{- end }}
+
```
torbsto, line 46: I think it could be a little weird to have the consumerGroup annotation on every resource (like Service and ScaledObject). Maybe we can instead split this into two functions, one that prints all annotations in `.Values.annotations` and one that uses it to build the whole block. Then, most resources can stay like you've implemented it and only the deployment would need to have a custom implementation.
Just as pseudocode to maybe illustrate it a bit:
```tpl

{{- define "streams-app.default-annotations" -}}
  {{- range $key, $value := .Values.annotations }}
    {{ $key | quote }}: {{ $value | quote }}
  {{- end }}
{{- end }}

{{- define "streams-app.annotations" -}}
{{- if  .Values.annotations }}
  annotations:
   {{- include "streams-app.annotations" . | nindent 4 }}
{{- end }}
{{- end }}
```

raminqaf: Good point! I have changed it. Here is the output of the deployment.yaml and service.yaml after template:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: bakdata-streams-app
  annotations:
    "MY_ANNOTATION": "foo-bar"
    "MY_ANNOTATION_2": "foo-bar-2"
    consumerGroup: "foo"
```

```yaml
apiVersion: v1
kind: Service
metadata:
  name: bakdata-streams-app
  annotations:
    "MY_ANNOTATION": "foo-bar"
    "MY_ANNOTATION_2": "foo-bar-2"
  labels:
    app: bakdata-streams-app
    chart: streams-app-0.1.0
    release: release-name
spec:
  selector:
    app: bakdata-streams-app
    release: release-name
  ports:
  type: ClusterIP
```

charts/streams-app/templates/_helpers.tpl
```diff
@@ -27,3 +27,31 @@ Create chart name and version as used by the chart label.
...
+{{/*
+Define annotations helper for Deployment.
+Includes default annotations and conditionally adds consumerGroup.
+Only includes annotations if there is content.
+*/}}
+{{- define "streams-app.deployment-annotations" -}}
+{{- include "streams-app.annotations" . }}
```
torbsto, line 50: There will be an error if you have a consumerGroup but no other annotation, I think

raminqaf: Should be fixed:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: bakdata-streams-app
  annotations:
    consumerGroup: "foo"
```

torbsto (approved):

PR #264, opened 2025-01-06
Replace kafka-junit with testcontainers

philipp94831:


streams-bootstrap-core/src/main/java/com/bakdata/kafka/util/ConsumerGroupClient.java
```diff
@@ -122,7 +133,11 @@ public Collection<ConsumerGroupListing> listGroups() {
      */
     public void deleteGroupIfExists(final String groupName) {
         if (this.exists(groupName)) {
-            this.deleteConsumerGroup(groupName);
+            try {
+                this.deleteConsumerGroup(groupName);
+            } catch (final GroupIdNotFoundException e) {
```
philipp94831, line 138: This is very weird. When executing tests, sometimes exists returns true, presumably because Kafka Streams resetter has reset offsets before on an non-existent consumer group. But when calling delete, the consumer group does not exist anymore. When doing some logging on between, it works as well but that is not a very robust solution. In the end, I think catching this exception does not hurt

streams-bootstrap-core/src/main/java/com/bakdata/kafka/util/ConsumerGroupClient.java
```diff
@@ -74,7 +75,12 @@ public void deleteConsumerGroup(final String groupName) {
         } catch (final InterruptedException ex) {
             Thread.currentThread().interrupt();
             throw new KafkaAdminException("Failed to delete consumer group " + groupName, ex);
-        } catch (final ExecutionException | TimeoutException ex) {
+        } catch (final ExecutionException ex) {
+            if (ex.getCause() instanceof RuntimeException) {
```
philipp94831, line 79: See comment below

torbsto: So all those checks are only required for `deleteGroupIfExists()`?

streams-bootstrap-core/src/testFixtures/java/com/bakdata/kafka/TestUtil.java
```diff
@@ -24,21 +24,13 @@
...
 public class TestUtil {
-    public static EmbeddedKafkaCluster newKafkaCluster() {
-        return provisionWith(newClusterConfig()
-                .configure(brokers()
-                        .with("transaction.state.log.num.partitions", 10)
-                        .with("offsets.topic.num.partitions", 10)
-                        .build())
-                .build());
+    public static ConfluentKafkaContainer newKafkaCluster() {
+        return new ConfluentKafkaContainer(DockerImageName.parse("confluentinc/cp-kafka:7.6.0"));
```
torbsto, line 34: Did you already look into the native image?

streams-bootstrap-core/src/testFixtures/java/com/bakdata/kafka/KafkaContainerHelper.java
```diff
@@ -0,0 +1,142 @@
...
+            final Map<String, Object> consumerConfig = new HashMap<>(this.properties);
+            consumerConfig.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG,
+                    KafkaContainerHelper.this.kafkaCluster.getBootstrapServers());
+            try (final Consumer<K, V> consumer = new KafkaConsumer<>(consumerConfig)) {
+                final List<PartitionInfo> partitionInfos = consumer.listTopics().get(output);
+                final List<TopicPartition> topicPartitions = partitionInfos.stream()
+                        .map(partition -> new TopicPartition(partition.topic(), partition.partition()))
+                        .collect(Collectors.toList());
+                consumer.assign(topicPartitions);
+                consumer.seekToBeginning(topicPartitions);
```
torbsto, line 136: Calling `seekToBeginning` here is not what I would expect when using this method. I sometimes have tests that include multiple cycles of producing + consuming/validating.

streams-bootstrap-core/src/testFixtures/java/com/bakdata/kafka/TestUtil.java
philipp94831: Apache image was first published for 3.7. Native image for 3.8. Will use them when upgrading

streams-bootstrap-core/src/testFixtures/java/com/bakdata/kafka/KafkaContainerHelper.java
philipp94831: Without using seek to beginning, nothing is read if topic already has messages. Also this mimics the behavior of the read method from old embedded Kafka

streams-bootstrap-core/src/main/java/com/bakdata/kafka/util/ConsumerGroupClient.java
philipp94831: I needed access to the cause of the execution exception. And for consistency I throw the cause for all execution exceptions

streams-bootstrap-core/src/testFixtures/java/com/bakdata/kafka/KafkaContainerHelper.java
torbsto: We could add `consumerConfig.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");` instead

philipp94831: I can do but is there any advantage over what we currently do? Also this is just test code

streams-bootstrap-core/src/main/java/com/bakdata/kafka/util/ConsumerGroupClient.java
torbsto: Maybe a retry in `deleteGroupIfExists` would be simpler? That keeps all the logic in there

philipp94831: I mean there is no reason to retry if we already know by the Exception type that the group does not exist. If another exception is thrown, the job would crash and at least Kubernetes would restart the pod

philipp94831: I feel like adding code for retry would make this much more complex without any benefit

streams-bootstrap-core/src/testFixtures/java/com/bakdata/kafka/KafkaContainerHelper.java
torbsto: Ah okay, I thought this is part of a library that can be user by others (me :smile:)

philipp94831: Yeah, in a next step that would be great (also with support for multi-broker clusters)

torbsto (approved):

PR #275, opened 2025-01-22
Call close on KafkaApplication only once

### LINKED ISSUE #263: Close method is called twice on shutdown
Once via shutdown hook and once via [manual call](https://github.com/bakdata/streams-bootstrap/blob/7abb6dafb6d91bbfc28b1dae1f6d1a7476419e46/streams-bootstrap-cli/src/main/java/com/bakdata/kafka/KafkaApplication.java#L300)

philipp94831:
Fixes #263


torbsto (commented):
I think a test would be helpful to ensure it is called only once

philipp94831:
> I think a test would be helpful to ensure it is called only once

I tried, the problem is that the shutdown hook is not invoked in the tests. We avoid system.exit with a junit plugin but this also avoids the shutdown hook. If we allow system.exit, Gradle tests are aborted

torbsto (approved):

PR #278, opened 2025-02-13
Retry topic verification

philipp94831:


streams-bootstrap-core/src/main/java/com/bakdata/kafka/util/TopicClient.java
```diff
@@ -55,6 +57,12 @@
 @Slf4j
 public final class TopicClient implements AutoCloseable {
 
+    private static final RetryConfig RETRY_CONFIG = RetryConfig.<Boolean>custom()
+            .retryOnResult(result -> result)
+            .failAfterMaxAttempts(false)
+            .maxAttempts(3)
+            .waitDuration(Duration.ofSeconds(5L))
```
yannick-roeder, line 64: Would it make sense to make all or some of these settings configurable?

yannick-roeder (changes requested):

streams-bootstrap-core/src/main/java/com/bakdata/kafka/util/TopicClient.java
philipp94831: I don't see a need for this atm. If it is later needed, we can add it

yannick-roeder (approved):

PR #280, opened 2025-02-21
Allow flexible configuration of JMX rules

philipp94831:


charts/producer-app/templates/deployment.yaml
```diff
@@ -12,6 +12,7 @@ metadata:
   labels:
     app: {{ template "producer-app.name" . }}
     chart: {{ template "producer-app.chart" . }}
+    streams-bootstrap/kind: {{ .Chart.Name }}
```
philipp94831, line 15: Any ideas for better names?

disrupted: looks fine to me

disrupted (approved):

PR #284, opened 2025-03-11
Create BOM

philipp94831:


streams-bootstrap-core/build.gradle.kts
```diff
@@ -20,7 +21,8 @@ dependencies {
     )
     implementation(group = "org.jooq", name = "jool", version = "0.9.15")
     implementation(group = "io.github.resilience4j", name = "resilience4j-retry", version = "1.7.1")
-    api(group = "com.bakdata.kafka", name = "error-handling-core", version = "1.6.1")
+    api(platform("com.bakdata.kafka:error-handling-bom:1.7.0"))
```
jkbe, line 24: I assume this can't be split into separate strings for `group`, `name` and `version`?

jkbe (approved):

streams-bootstrap-core/build.gradle.kts
philipp94831: Unfortunately not. Once we migrate to version catalog it works though

PR #300, opened 2025-04-08
Add example for version plugin

torbsto:


build.gradle.kts
```diff
@@ -1,8 +1,29 @@
...
+versionCatalogUpdate {
+    sortByKey.set(false)
+}
+
+fun isNonStable(version: String): Boolean {
+    val stableKeyword = listOf("RELEASE", "FINAL", "GA").any { version.uppercase().contains(it) }
+    val regex = "^[0-9,.v-]+(-r)?$".toRegex()
+    val isStable = stableKeyword || regex.matches(version)
+    return isStable.not()
 }
```
torbsto, line 27: This could be moved to a gradle plugin

gradle/libs.versions.toml
torbsto, line 1: This is the output of running `versionCatalogUpdate`. The formatting is rather opinionated but I don't mind because you don't really touch the file anymore

torbsto, line 1: This is the output of running versionCatalogUpdate. The formatting is rather opinionated but I don't mind because you don't really touch the file anymore

philipp94831:
@torbsto I think with dependabot running less frequently, we don't need this. Thoughts?

PR #315, opened 2025-05-13
Separate schema registry from TestTopologyFactory

philipp94831:


streams-bootstrap-test/build.gradle.kts
```diff
@@ -12,6 +13,12 @@ dependencies {
     testRuntimeOnly(libs.junit.platform.launcher)
     testImplementation(libs.junit.jupiter)
     testImplementation(libs.assertj)
+    testImplementation(libs.kafka.streams.avro.serde) {
+        exclude(group = "org.apache.kafka", module = "kafka-clients") // force usage of OSS kafka-clients
+    }
+    testImplementation(libs.kafka.protobuf.provider) {
+        exclude(group = "org.apache.kafka", module = "kafka-clients") // force usage of OSS kafka-clients
+    }
```
jkbe, line 21: Doesn't this exclude the OSS kafka-clients rather than force their usage?

jkbe: Ah ok, the comment sounded like there was a distinction between org.apache.kafka.kafka-clients and another non-OSS client library under io.confluent namespace. Thanks.

jkbe (approved):

PR #338, opened 2025-07-03
Convert start application methods to instance methods

philipp94831:


raphala (commented):
Looks good otherwise.

PR #354, opened 2025-07-17
Add Github Workflow for Github Pages Wiki

vostres:


philipp94831 (changes requested):

.github/workflows/build-and-publish.yaml
```diff
@@ -22,3 +22,69 @@ jobs:
...
+        run: poetry install --no-root
+
+      - name: Build docs
+        run: poetry run mkdocs build -f docs/mkdocs.yml
+
+  publish-docs-from-main:
+    name: Publish Docs (main)
+    runs-on: ubuntu-24.04
+    needs: docs
+    if: ${{ github.ref == 'refs/heads/master' }}
```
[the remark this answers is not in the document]
vostres: check, tested in playground

.github/workflows/build-and-publish.yaml
```diff
@@ -22,3 +22,69 @@ jobs:
       ossrh-username: ${{ secrets.SONATYPE_OSSRH_USERNAME }}
       ossrh-password: ${{ secrets.SONATYPE_OSSRH_PASSWORD }}
       github-token: ${{ secrets.GH_TOKEN }}
+  docs:
```
philipp94831, line 25: Maybe we should add this to our java-library workflow 🤔 then we can reuse it in other repos and we can restructure the workflow a bit 
image
I think it makes sense to have the "Build Docs" job run in parallel to the java build and then the publish docs job depends on assess code quality and "Build docs"
image

philipp94831: Ideally we have an input param build-docs which is false by default

docs/justfile
torbsto, line 1: Do we want to keep the justfile? It seems pretty useless and is mainly a leftover of Quick's setup that we've also used for kpops

docs/mkdocs.yml
torbsto, line 1: I'd suggest to move the mkdocs.yml to the root so that we don't need to keep the actual docs in the directory `/docs/docs`. This is also something that was copied from Quick but is not really needed anymore I think

torbsto (commented):
@disrupted I guess the setup is mostly copied from kpops, but the setup there seems to be copied from Quick. Maybe we can evaluate if we want to simplify this by 1) removing the justfile and 2) move the mkdocs.yaml to the root so that documentation can live in `/docs` instead of `/docs/docs`. Those were two decisions that worked well for Quick, but seem unnecessary to me for our projects nowadays

.github/workflows/build-and-publish.yaml
```diff
@@ -22,3 +22,64 @@ jobs:
...
```
philipp94831, line 25: newline before pls

docs/docs/index.md
```diff
@@ -0,0 +1 @@
+# Hello World 🚀
```
philipp94831, line 1: Coming soon

PR #356, opened 2025-07-21
Host Javadoc on Github Pages

vostres:


docs/docs/javadoc.md
```diff
@@ -0,0 +1,10 @@
+# Javadoc
+
+Choose a module to view its generated Javadoc:
+
+- [streams-bootstrap-cli](../javadoc/streams-bootstrap-cli/index.html){target=_blank}
```
philipp94831, line 5: Is it possible to build a combined JavaDoc instead? https://stackoverflow.com/questions/63287292/how-to-combine-multiple-javadoc-into-one-using-gradle

build.gradle.kts
```diff
@@ -13,8 +14,18 @@ allprojects {
         maven(url = "https://packages.confluent.io/maven/")
         maven(url = "https://central.sonatype.com/repository/maven-snapshots")
     }
+
+    dependencies {
+        subprojects {
+            plugins.withId("java") {
+                javadoc(this@subprojects)
+                javadocClasspath(libs.lombok)
```
philipp94831, line 22: Please don't use version catalog here but
`javadocClasspath("org.projectlombok:lombok:${LombokExtension.LOMBOK_VERSION}")`

vostres: Did you add any additional dependencies? `LombokExtension` is not resolving either locally or in the workflow.

docs/docs/javadoc.md
```diff
@@ -0,0 +1,5 @@
+# Java API Reference
+
+The complete Java API documentation for **Streams Bootstrap** is available below:
+
+- [View Javadoc](all/index.html){target=_blank}
```
philipp94831, line 5: I don't think we need this page. Can the menu bar just link to the docs directly?

.github/dependabot.yaml
```diff
@@ -30,6 +30,7 @@ updates:
           - "com.bakdata.release"
           - "com.bakdata.sonar"
           - "com.bakdata.sonatype"
+          - "io.freefair*"
```
philipp94831, line 33: Please create a separate group for io.freefair

docs/docs/javadoc.md
vostres: I tried to embed script which opens a new tab to javadoc.md. locally it works, but not on github pages. alternatively I can just link javadoc, but then it will not open a new tab. do you have a preference?

philipp94831: I think we just get rid of the new tab thing then. Ideally we don't do any javascript shenanigans

.github/workflows/build-and-publish.yaml
```diff
@@ -74,7 +118,7 @@ jobs:
           filters: |
             docs:
               - added|deleted|modified: 'docs/**'
-
+
```
philipp94831, line 121: Revert

PR #380, opened 2025-09-15
Provide lineage information via headers

philipp94831:


streams-bootstrap-core/src/main/java/com/bakdata/kafka/streams/kstream/KTableX.java
```diff
@@ -429,4 +429,55 @@ <VR, KO, VO> KTableX<K, VR> leftJoin(KTable<KO, VO> other, BiFunction<K, V, KO>
     <VR, KO, VO> KTableX<K, VR> leftJoin(KTable<KO, VO> other, BiFunction<K, V, KO> foreignKeyExtractor,
             ValueJoiner<V, VO, VR> joiner, TableJoined<K, KO> tableJoined,
             MaterializedX<K, VR, KeyValueStore<Bytes, byte[]>> materialized);
+
```
torbsto, line 432: What is the use case for adding lineage to a KTable?

philipp94831: You can create a table directly via the StreamsBuilder. And because there is also the transformer API available, I added it for completeness

PR #385, opened 2025-10-31
Add consumer application

raphala:


streams-bootstrap-core/src/main/java/com/bakdata/kafka/streams/kstream/KStreamXImpl.java
torbsto: Why are the changes in this file necessary? Maybe it would make sense to move them into their own PR?

raphala: Good catch, this is part of the [update to Kafka 4.1](https://github.com/bakdata/streams-bootstrap/pull/377)

streams-bootstrap-cli/src/main/java/com/bakdata/kafka/mixin/ConsumerOptions.java
```diff
@@ -0,0 +1,46 @@
...
+@Setter
+@NoArgsConstructor
+public class ConsumerOptions {
+    @CommandLine.Option(names = "--volatile-group-instance-id", arity = "0..1",
+            description = "Whether the group instance id is volatile, i.e., it will change on a Streams shutdown.")
+    private boolean volatileGroupInstanceId;
+    @CommandLine.Option(names = "--application-id",
+            description = "Unique application ID to use for Kafka Streams. Can also be provided by implementing "
+                    + "StreamsApp#getUniqueAppId()")
+    private String applicationId;
```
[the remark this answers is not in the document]
raphala: I tried to reuse this Mixin, but you're right. Only Kafka Streams apps should have an 'application-id' option.

streams-bootstrap-core/src/main/java/com/bakdata/kafka/streams/StreamsCleanUpConfiguration.java
```diff
@@ -84,4 +86,30 @@ void runResetHooks() {
...
+     * registered clean and topic hooks.
+     *
+     * @return {@link ProducerCleanUpConfiguration} instance
+     */
+    public ProducerCleanUpConfiguration toProducerCleanUpConfiguration() {
+        final ProducerCleanUpConfiguration configuration = new ProducerCleanUpConfiguration();
+        this.cleanHooks.forEach(configuration::registerCleanHook);
+        this.topicHooks.forEach(configuration::registerTopicHook);
+        return configuration;
+    }
```
jkbe: These two methods seem to be unused. Does it make sense to keep them around regardless?

streams-bootstrap-core/src/main/java/com/bakdata/kafka/consumer/ConfiguredConsumerApp.java
```diff
@@ -0,0 +1,144 @@
...
+ * @param <T> type of {@link ConsumerApp}
+ */
+public record ConfiguredConsumerApp<T extends ConsumerApp>(@NonNull T app,
+                                                           @NonNull ConsumerAppConfiguration configuration)
+        implements ConfiguredApp<ExecutableConsumerApp<T>> {
+    private static Map<String, Object> createBaseConfig() {
+        final Map<String, Object> kafkaConfig = new HashMap<>();
+
+        kafkaConfig.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
+        kafkaConfig.put(ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG, "false");
```
jkbe: Is the goal here to enable exactly-once semantics? If yes, we should add `isolation.level: read_committed` so that we don't read any uncommitted records from upstream Streams apps: https://docs.confluent.io/platform/current/installation/configuration/consumer-configs.html#isolation-level

streams-bootstrap-core/src/main/java/com/bakdata/kafka/consumer/ConsumerCleanUpRunner.java
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    @RequiredArgsConstructor
+    private class Task {
+
+        private final @NonNull AdminClientX adminClient;
+
+        private void reset() {
+            final Optional<ConsumerGroupDescription> groupDescription =
+                    this.adminClient.consumerGroups().group(ConsumerCleanUpRunner.this.groupId).describe();
```
jkbe: You could extract a variable for `this.adminClient.consumerGroups().group(ConsumerCleanUpRunner.this.groupId)` and reuse below. Or is it possible that the group changes in between the two statements that use it?

streams-bootstrap-core/src/main/java/com/bakdata/kafka/consumer/ConsumerExecutionOptions.java
```diff
@@ -0,0 +1,90 @@
...
+ * Options to run a Kafka Consumer app
+ */
+@Builder
+public final class ConsumerExecutionOptions {
+
+    /**
+     * Hook that is called after the {@link ConsumerRunnable} is started
+     */
+    @Builder.Default
+    private final @NonNull java.util.function.Consumer<RunningConsumer> onStart = runningStreams -> {};
```
jkbe:
```suggestion
    private final @NonNull java.util.function.Consumer<RunningConsumer> onStart = runningConsumer -> {};
```

jkbe (approved):

charts/consumer-app/templates/deployment.yaml
```diff
@@ -0,0 +1,254 @@
...
+              {{- if .readOnly }}
+              readOnly: true
+              {{- end }}
+              {{- if .subPath}}
+              subPath: {{.subPath }}
+              {{- end }}
+            {{- end }}
+            {{- if and .Values.persistence.enabled .Values.statefulSet }}
+            - name: datadir
+              mountPath: /tmp/kafka-consumer
```
philipp94831: @raphala What is the purpose of this? For Kafka Streams there is a dedicated temp dir which is used for persistent state stores. However, here it seems unnecessary?

PR #447, opened 2026-05-04
feat(kip-1071): implement reset for new streams groups

jkbe:


streams-bootstrap-test/src/main/java/com/bakdata/kafka/ConsumerGroupVerifier.java
```diff
@@ -40,19 +41,29 @@
...
- * Utility class to verify the state of a Kafka consumer group
+ * Utility class to verify the state of a Kafka consumer group or streams group
  */
 @Slf4j
 @RequiredArgsConstructor
 public class ConsumerGroupVerifier {
 
     private final @NonNull String group;
     private final @NonNull Supplier<AdminClientX> adminClientSupplier;
+    private final boolean streamsGroupProtocol;
```
philipp94831, line 58: Rather than havin this flag, we should have two implementations of an interface that you can plug in here

PR #451, opened 2026-06-05
Upgrade to Kafka 4.3

philipp94831:


streams-bootstrap-core/src/main/java/com/bakdata/kafka/producer/ConfiguredProducerApp.java
```diff
@@ -36,7 +36,7 @@
 import lombok.NonNull;
 import lombok.RequiredArgsConstructor;
 import org.apache.kafka.clients.producer.ProducerConfig;
-import org.apache.kafka.common.record.CompressionType;
+import org.apache.kafka.common.record.internal.CompressionType;
```
torbsto, line 39: Do you know why they've made it internal? Are there any alternatives?

torbsto (approved):

streams-bootstrap-core/src/main/java/com/bakdata/kafka/producer/ConfiguredProducerApp.java
philipp94831: No clue. Using `Compression.gzip().build().type().toString()` instead of `CompressionType.GZIP.toString()` does not seem better to me

PR #472, opened 2026-07-31
Add explicit KEDA triggers and scaling modifiers

raphala:
Allow apps to replace generated Kafka lag triggers with explicit KEDA triggers and configure scaling through scalingModifiers. Existing additionalTriggers behavior remains compatible.

### DIFF
--- a/charts/kafka-app/templates/_scaled-object.yaml
+++ b/charts/kafka-app/templates/_scaled-object.yaml
@@ -1,8 +1,20 @@
 {{- define "kafka-app.scaled-object" -}}
 {{ if .Values.autoscaling.enabled }}
   {{- $uniqueId := coalesce .Values.kafka.applicationId .Values.kafka.groupId -}}
-  {{- if not (and $uniqueId .Values.autoscaling.lagThreshold) }}
-  {{- fail "When autoscaling is enabled, you must set both .Values.kafka.applicationId (or .Values.kafka.groupId) and .Values.autoscaling.lagThreshold" }}
+  {{- $explicitTriggers := .Values.autoscaling.triggers | default list -}}
+  {{- $additionalTriggers := .Values.autoscaling.additionalTriggers | default list -}}
+  {{- $scalingModifiers := .Values.autoscaling.scalingModifiers | default dict -}}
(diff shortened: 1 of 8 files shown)


charts/kafka-app/templates/_scaled-object.yaml
```diff
@@ -1,8 +1,23 @@
...
-  {{- if not (and $uniqueId .Values.autoscaling.lagThreshold) }}
-  {{- fail "When autoscaling is enabled, you must set both .Values.kafka.applicationId (or .Values.kafka.groupId) and .Values.autoscaling.lagThreshold" }}
+  {{- $explicitTriggers := .Values.autoscaling.triggers | default list -}}
+  {{- $additionalTriggers := .Values.autoscaling.additionalTriggers | default list -}}
+  {{- $scalingModifiers := .Values.autoscaling.scalingModifiers | default dict -}}
+  {{- $hasTopicSources := or .Values.kafka.inputTopics .Values.autoscaling.internalTopics .Values.autoscaling.topics .Values.kafka.labeledInputTopics -}}
+  {{- if and $explicitTriggers $additionalTriggers }}
+  {{- fail "autoscaling.triggers and autoscaling.additionalTriggers cannot be used together" }}
+  {{- end }}
+  {{- if and $scalingModifiers (not $explicitTriggers) }}
```
philipp94831, line 11: why does it require explicit triggers?

raphala: KEDA scaling formulas reference triggers by name, but the chart generated triggers currently have no names. For now, I think it makes sense to require explicit triggers instead of generating names.

philipp94831: Can't we name each trigger by using the topic name? That should be unambigous

raphala: Yes, but we would have to sanitize it, as hyphens would interfere with the `scalingModifiers` for instance. We could replace them with underscores, but then we could cause collisions.

philipp94831: Then lets skip it for now. But would be nice if we can expand it soon

philipp94831 (approved):

